Issue #77, opened 2019-09-04
Tabs are not shown


prizrak9:
my old tabs like newly created ones are not displayed in the bar.
![image](https://user-images.githubusercontent.com/26070759/64227139-efa8d880-ceea-11e9-92fc-81e981fbf038.png)

prizrak9:
Oh, it just needed to reopen the bar. Restart of the browser had no effect tho.

prizrak9:
After another browser restart all broke again and this time F1 won't help. That pinned tab is shown for some reason but even if I unpin it, it doesn't go anywhere(it is truly unpinned) and the option "unpin" still present.

mbnuqw:
Looks like upgrading issue. Can you screenshot 'Storage' section in the debug page?
Thanks.

mbnuqw:
Also you can backup addon data (Setting/Help/Export data) and try to reset settings.

shokinn:
Hey, same Problem here.

Here is the requested screenshot.
![Screenshot 2019-09-04 at 08 46 15](https://user-images.githubusercontent.com/10045300/64231673-9348b780-cef0-11e9-840d-d1d7338f58db.png)

Resetting the addon settings works (at least for myself).
After importing it, it's breaking again. So it looks like it is some settings issue.

mbnuqw:
@shokinn, thanks for the info and sorry for that bug.

berdon:
![image](https://user-images.githubusercontent.com/898462/64290414-acbb2500-cf2b-11e9-9bba-8ad249574b57.png)

berdon:
```
  if ('prev_act' === e.activateAfterClosing) {
    let i;
    for (let o = e.actTabs.length; o--; ) if (i = e.actTabs[o], (!r || !r.includes(i)) && i !== t.id && e.tabsMap[i]) {
      n = e.tabsMap[i];
      break
    }
  }
```

e.actTabs is null

berdon:
**Workaround**
- Change "After closing current tab activate" to none
- Close Sidebery (hit the x)
- Re-open Sidebery

mbnuqw:
Thank you all and @berdon especially.
Fixed in v3.0.3

PR #79, opened 2019-09-04
#77 - Handling an empty actTabs gracefully

berdon:
state.actTabs doesn't look like it ever gets created unless `onTabActivated` is called. I'm not sure why that's not happening but this at least reverts to the "previous" tab until you have time to look into it.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -329,7 +329,7 @@ function findSuccessorTab(state, tab, exclude) {
   }
 
   // Previous tab
-  if (state.activateAfterClosing === 'prev' || (state.actTabs == null && state.activateAfterClosing === 'prev_act')) {
+  if (state.activateAfterClosing === 'prev') {
     for (let i = tab.index, prev; i--; ) {
       prev = state.tabs[i]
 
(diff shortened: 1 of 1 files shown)


mbnuqw:
Thanks for this PR! I suggest to just change only line 384 to 
`if (state.actTabs && state.activateAfterClosing === 'prev_act') {`
In this case, if `state.actTabs` is uninitialized Sidebery will do nothing, and Firefox will use default action.

src/utils.js
```diff
@@ -329,7 +329,7 @@ function findSuccessorTab(state, tab, exclude) {
   }
 
   // Previous tab
-  if (state.activateAfterClosing === 'prev') {
+  if (state.activateAfterClosing === 'prev' || (state.actTabs == null && state.activateAfterClosing === 'prev_act')) {
```
mbnuqw: So, here we should leave `if (state.activateAfterClosing === 'prev') {`

src/utils.js
```diff
@@ -381,7 +381,7 @@ function findSuccessorTab(state, tab, exclude) {
   }
 
   // Previously active tab
-  if (state.activateAfterClosing === 'prev_act') {
+  if (state.actTabs != null && state.activateAfterClosing === 'prev_act') {
```
mbnuqw: And here we just add the actTabs check: `if (state.actTabs && state.activateAfterClosing === 'prev_act') {`

PR #540, opened 2021-12-19
Fixes #539 - Don't use the forbidden text/x-moz-text-internal

### LINKED ISSUE #539: Some DataTransfer setData calls throw in Nightly
## Steps to reproduce

Drag and drop a tab. Probably also a bookmark?

## Expected behavior

It works

## Actual behavior

It doesn't work, because some of the [setData](https://github.com/mbnuqw/sidebery/search?q=setData) calls that include `x-moz` like `e.dataTransfer.setData('text/x-moz-text-internal', this.tab.url)` now throw. `text/x-moz-url` should continue to be allowed.

@mbnuqw This requires changes from your side. For now just wrapping the individual setData calls in try..catch would work. Or maybe just removing them, but I am not sure if they are necessary.

## Environment info
OS:  All
Firefox version:  Firefox Nightly (97)
Addon version:  4.9.4

evilpie:
I think this isn't necessary, because text/plain should be able to
replace it.

### DIFF
--- a/src/sidebar/components/bookmark.vue
+++ b/src/sidebar/components/bookmark.vue
@@ -213,7 +213,6 @@ export default {
       State.dragXStart = e.clientX
 
       // Set drag info
-      e.dataTransfer.setData('text/x-moz-text-internal', this.node.url)
       e.dataTransfer.setData('text/uri-list', this.node.url)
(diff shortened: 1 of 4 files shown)


mbnuqw:
Thank you!

PR #605, opened 2022-04-14
Fix drag & drop with Proton theme and custom CSS

daniel-stoneuk:
Custom CSS changes such as `--tabs-height` were not applied to Sidebar's element sizes when using the Proton theme.

Other themes were unaffected if their CSS was loaded before the 500 debounced call to `recalcElementSizes`. 

This change will recalculate element sizes if custom CSS is used and will not affect performance as the call is debounced.

### Changes

* Remove unnecessary .removeAttribute method call
* Recalculate element sizes when custom CSS changes


mbnuqw:
Thank you!

Issue #678, opened 2022-07-01
various issues I've found first time using this


br4nnigan:
## Description
- "undo close tab" should restore the tab at its original position, not at the very bottom
- there should be a tab context menu item for closing the whole tree vs closing the parent only, currently I'm using the option to close the tree _when it's folded_ as an alternative
- reopen in/move to NEW PANEL
- the option to disable title ("hover tooltip") would be appreciated, or even optionally sanitized titles (query params stripped)
- why is "flatten" disabled when the tree is not collapsed?
- move to GROUP

dexeonify:
> "undo close tab" should restore the tab at its original position, not at the very bottom

Works for me.

> there should be a tab context menu item for closing the whole tree vs closing the parent only

If you **right-click** on the parent tab's favicon, it will select the whole tree and display a context menu. From there, you can click "Close" as usual. Alternatively, you can hover over the parent tab and **right-click** the close button (x), this will also close the whole tree.

In v5, there are context menu items for "Close" (selected tabs), "Close descendants" and "Close branch". You can have a field day with these :)

> reopen in/move to NEW PANEL

This is added in v5.

> or even optionally sanitized titles (query params stripped)

Should be a job for another extension. uBlock Origin has this functionality too.

> why is "flatten" disabled when the tree is not collapsed?

Maybe because it only works if you select the whole tree. You can add keybinds to indent/outdent the child tabs.

> move to GROUP

See #619

br4nnigan:
> > "undo close tab" should restore the tab at its original position, not at the very bottom
> 
> Works for me.

Ok seems to be an edge case, I can't reproduce.

> > there should be a tab context menu item for closing the whole tree vs closing the parent only
> 
> If you **right-click** on the parent tab's favicon, it will select the whole tree and display a context menu. From there, you can click "Close" as usual. Alternatively, you can hover over the parent tab and **right-click** the close button (x), this will also close the whole tree.
> In v5, there are context menu items for "Close" (selected tabs), "Close descendants" and "Close branch". You can have a field day with these :)

very good to know, thanks

> > reopen in/move to NEW PANEL
> 
> This is added in v5.

👍

> > or even optionally sanitized titles (query params stripped)
> 
> Should be a job for another extension. uBlock Origin has this functionality too.

I dont mean modifying the document title, but the title attribute of the TAB, where sidebery (immitating native behavior) adds the URL. I'm gonna take a look into the code, maybe I can create a PR

[...]

PR #681, opened 2022-07-03
add option to ommit/shorten url in tab title attribute (hover tooltip)

br4nnigan:
as I said [here](https://github.com/mbnuqw/sidebery/issues/678) I'd like to have an option for the tab titles - not the actual text content which can be altered separetely by other addons but the title attribute within the sidebery DOM which is used for the hover tooltip.

seems to be working fine, however some translations are still missing


mbnuqw:
Thanks for this PR, works great.

PR #743, opened 2022-08-31
Added: German language

Qjo1:
Integrated full German translation.

While testing to make sure I used the appropriate translations everywhere, I noticed some deviations in the dict.common file, as I described in discussion #726. I ended up simply providing shortened translations, which work well for both, the menus and the settings.


mbnuqw:
Thank you!

Issue #752, opened 2022-09-09
Middle click on "new tab" button should open child tab beneath the active one


emvaized:
## Description
Currently you can click "New tab" button on the bottom of tabs list to open new level 0 tab, and you can right click to choose the container. Would be great if you could also middle click on button to open new child tab, e.g. you're currently on level 0 tab, and it will open level 1 tab beneath it.

dexeonify:
<kbd>Ctrl</kbd>+Left click does that. Probably one of Sidebery hidden mouse shortcuts. Would be nice if we can define shortcuts for <kbd>Shift</kbd>+Left click or middle click new tab too.

emvaized:
Thanks, I know about `Ctrl` + Left click. My point is that it would be great to have it on middle click as well, so that it won't require to reach out for keyboard. 
It also makes sense since everywhere through out the system middle click usually behaves as `Ctrl` + Left click.

emvaized:
It seems that in current version middle click on 'new tab' button calls "[reopen()](https://github.com/emvaized/sidebery/blob/a28eec1bd8f48039dabd95e72701b8ca23582e7e/src/sidebar/components/bar.new-tab.vue#L133)" function, which, I guess by name, should reopen last closed tab, or reopen current tab. 

But in my case this action just does nothing. 
@dexeonify Does it do anything for you, when you middle click on 'new tab' button?

dexeonify:
> But in my case this action just does nothing. @dexeonify Does it do anything for you, when you middle click on 'new tab' button?

No, which is why I suggested:
> Would be nice if we can define shortcuts for <kbd>Shift</kbd>+Left click or middle click new tab too.

But now that you mention it, I do remember a commit regarding middle-clicking and alt-clicking the new tab button. I can't seem to find it now.

dexeonify:
Ah yes, it's these two commits: mbnuqw/sidebery@0aef672 and mbnuqw/sidebery@7a13a06. Both don't seem to work, might be a regression somewhere.

emvaized:
I think, middle click with no modifiers on this button should create new child tab, as `Ctrl` + Left click is already assigned to do it, and it is common convention in all apps including Firefox that middle click should behave the same as `Ctrl` + Left click. 

And maybe, should use `Alt` + Left click to reopen tab?

emvaized:
I can create pull request about it. But non-working `reopen()` function would be another thing to deal with.

mbnuqw:
[...]

PR #753, opened 2022-09-09
Middle click on 'new tab' button to open new child tab

emvaized:
Since `Ctrl` + Left click action on new tab bar is hard coded to perform "Open new child tab", I find it more appropriate to remap middle click action to perform "Open child tab" action as well, as it's common convention in all apps, including Firefox, that middle click should behave the same as `Ctrl` + Left click.

In order to not loose current "reopen" functionality, I remapped it to `Ctrl` + Middle click, and also as `Alt` + Left click, as it used to [behave](https://github.com/mbnuqw/sidebery/commit/7a13a06) this way before the regression.

Associated issue: https://github.com/mbnuqw/sidebery/issues/752

### DIFF
--- a/src/sidebar/components/bar.new-tab.vue
+++ b/src/sidebar/components/bar.new-tab.vue
@@ -121,6 +121,11 @@ function onNewTabMouseDown(e: MouseEvent, btn?: NewTabBtn): void {
       return
     }
 
+    if (e.altKey) {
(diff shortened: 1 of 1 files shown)


mbnuqw:
Thanks for the PR. As I noted in the issue I'll add setting for middle click behavior. (and will release b26 soon)

PR #789, opened 2022-10-07
Add titles to tab buttons

ykrx:
Added titles to the buttons on the group page.

button-hover-titles_AdobeExpress

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "sidebery",
-  "version": "5.0.0b26",
+  "version": "5.0.0b27",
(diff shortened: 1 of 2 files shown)


mbnuqw:
Thanks for the PR. There are a few issues with it:
- package-lock.json shouldn't be changed
- Reordered import statements
- No localization (see [url page](https://github.com/mbnuqw/sidebery/blob/v5/src/page.url/url.ts#L103) [example](https://github.com/mbnuqw/sidebery/blob/v5/src/_locales/dict.browser.json#L401))

Issue #831, opened 2022-11-03
Option to hide the header of a folded tab group


masoncowen:
## Description
Currently the head tab of a group stays shown if you fold the group and have "Hide folded tab groups" enabled.
Could we get the option to also hide the header tab when it is folded?

PR #872, opened 2022-11-30
Create a "hide group tabs"

sarchar:
The hide groups option will hide group tabs when the group is folded and not selected.

Useful for windows that have a lot of tab groups, as the tab group pages tend to clutter the the tabs bar.

### DIFF
--- a/src/_locales/dict.setup-page.ts
+++ b/src/_locales/dict.setup-page.ts
@@ -507,6 +507,12 @@ export const setupPageTranslations: Translations = {
     de: 'Native Scrollleiste verwenden',
     zh_CN: '使用原生滚动条',
   },
+  'settings.hide_group_tabs': {
+    en: 'Hide group tabs of collapsed tabs',
+    ru: '',
+    de: '',
+    zh_CN: '',
+  },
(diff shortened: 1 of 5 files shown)


sarchar:
Hey everyone,

I have a LOT of tab groups, and thus a lot of tab group pages. I have found that it's nearly impossible to navigate using the top tabs bar of the browser window when it's 90% tab group pages.

I believe this new feature will be quite useful to others but my implementation is probably incomplete. There may be many, many edge cases that don't properly hide the group tabs.  Please let me know and I will try to cover those as they're discovered.

Also, since I'm not familiar at all with the code base, I'm sure I missed out on making use of the proper API to do things. In particular, I think tab.vue could be improved by using a deselect tab event/callback but I'm unaware of any that exist already.

Cheers,
Sarchar

sarchar:
To demonstrate the difference, here is before:

image

And after with tabs collapsed:

image

mbnuqw:
Thank you, this PR has a few issues, but it's a good start, I'll merge it and complete the implementation.

This feature was requested in #831 (just a mention)

Issue #660, opened 2022-06-13
Include rule not working


andinus:
## Steps to reproduce

- Add 'github.com' to Include URLs for a container.
- Grant necessary permission (Accessing all website data).
- Navigate to github.com.
- Opens in default container.

## Expected behavior

- github.com should switch to chosen container.

### Addon data
<details><code>
{
  "settings": {
    "version": "4.10.1",
    "nativeScrollbars": false,
    "selWinScreenshots": true,
    "tabsCheck": true,
    "tabsFix": "notify",
    "markWindow": false,
    "markWindowPreface": "[Sidebery] ",
    "ctxMenuNative": false,
    "autoHideCtxMenu": "none",
    "ctxMenuRenderInact": true,
    "ctxMenuIgnoreContainers": "",
    "navBarLayout": "horizontal",
    "navBarInline": false,
    "hideAddBtn": false,
    "hideSettingsBtn": false,
    "navBtnCount": true,
    "hideEmptyPanels": true,
    "navActTabsPanelLeftClickAction": "none",
    "navActBookmarksPanelLeftClickAction": "none",
    "navMidClickAction": "none",
    "navSwitchPanelsWheel": true,
    "groupLayout": "list",
    "skipEmptyPanels": false,
    "dndTabAct": true,
    "dndTabActDelay": 750,
    "dndTabActMod": "none",
    "dndExp": "pointer",
    "dndExpDelay": 0,
    "dndExpMod": "none",
    "stateStorage": "global",
    "warnOnMultiTabClose": "collapsed",
    "activateOnMouseUp": true,
    "activateLastTabOnPanelSwitching": true,
    "showTabRmBtn": true,
    "showTabCtx": true,
    "hideInact": false,
    "activateAfterClosing": "next",
    "activateAfterClosingPrevRule": "visible",
    "activateAfterClosingNextRule": "tree",
    "activateAfterClosingGlobal": false,
    "activateAfterClosingNoFolded": true,
    "activateAfterClosingNoDiscarded": true,
    "shiftSelAct": true,
    "askNewBookmarkPlace": false,
    "tabsRmUndoNote": true,
    "nativeHighlight": false,
    "tabsUnreadMark": false,
    "tabsReloadLimit": 5,
    "tabsReloadLimitNotif": true,
    "tabsPanelSwitchActMove": true,
    "moveNewTabPin": "start",
    "moveNewTabParent": "last_child",
    "moveNewTabParentActPanel": false,
[...]

PR #893, opened 2023-01-10
Add information for workaround. Relates to #660

HT43-bqxFqB:
I do not know much about Firefox addons, but this info would have helped me a lot. This does not fix the missing permissions workflow in #660 but it guides the user.


mbnuqw:
Thanks

PR #895, opened 2023-01-13
Fixed: Switched panel,  Tabs.activeId not update.

SLin0218:
```
function onMouseUp(e: MouseEvent): void {
  ...
  if (
    Settings.state.tabsSecondClickActPrev &&
    // result = true,  first switch panel then click tab. will jump back prev panel tab.   
    props.tab.id === Tabs.activeId &&
    withoutMods &&
    !activating
  ) {
    const history = Tabs.getActiveTabsHistory()
    const prevTabId = history.actTabs[history.actTabs.length - 1]
    if (prevTabId !== undefined) browser.tabs.update(prevTabId, { active: true })
  }
  ...
}
```

### DIFF
--- a/src/services/tabs.fg.actions.ts
+++ b/src/services/tabs.fg.actions.ts
@@ -1336,6 +1336,7 @@ export function activateLastActiveTabOf(panelId: ID): void {
   if (!tab || tab.panelId !== p.id) tab = panelTabs[0]
   if (tab && tab.discarded) tab = panelTabs.find(t => !t.discarded)
   if (tab) {
+    Tabs.activeId = tab.id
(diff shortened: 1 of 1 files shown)


mbnuqw:
It's intended behavior - currently, backward activation (tab flip) use the global history (`Tabs.getActiveTabsHistory()` called without `panelId` argument). But I can add a sub-option to limit tab-flip switching to panel-only. You can open a feature request to track or implement it yourself in new PR.

If I misunderstood the initial issue, please, provide steps to reproduce and your settings (from debug info in the Help section).

SLin0218:
I understand, thanks, i try open a new PR.

PR #899, opened 2023-01-16
Added: Backward activation when clicking on the active tab. (Only pan…

SLin0218:
Added: Backward activation when clicking on the active tab. (Only panel sub option)


SLin0218:
#895

mbnuqw:
Thank you!

PR #972, opened 2023-03-04
Add option to show titles of globally-top pinned tabs

alan-palacios:
The option was only available for `in panel - top` position. Now it has the same style in form of list rather than grid when `Show titles of pinned tabs` is on.


mbnuqw:
Thanks!

Issue #780, opened 2022-09-29
Inconsistent / confusing middle-click behaviour on Bookmarks & History.


guigirl42:
## Steps to reproduce

1. Enable Bookmarks & History in the Navigation bar.
2. Make Settings as per pic's yellow highlight 
![image](https://user-images.githubusercontent.com/76090786/192971883-35ca1edd-81b8-4fae-a3d6-cccf1b1d0597.png)

3. Change from active Tabs panel to Bookmarks panel.
4. Find one or many bookmarks, then one by one middleclick on it/them.
5. Change back to the Tabs panel & observe if any bookmarks were opened there.
6. Change to History panel.
7. Find one or many old tabs, then one by one middleclick on it/them.
8. Change back to the Tabs panel & observe if any recovered tabs were opened there.
9. Change to History panel again.
10. Find one or many old tabs, then one by one leftclick on it/them.
11. Change back to the Tabs panel & observe if any recovered tabs were opened there.

## Expected behaviour

In each case, the requisite BMs or old tabs should have opened in the Tabs panel.

## Actual behaviour
- At point5, _many many_ times **nothing** will have been opened, but _sometimes_, inexplicably, the operation will have worked perfectly. I've been confused by this inconsistency / unpredictably for weeks [ie, spanning Betas 25, 26, 27]. 
- At point8, always, **nothing** will have been opened.
- At point11, always[?], **the correct tabs** will have been opened.

### Addon data
<details><code>
{
  "settings": {
    "nativeScrollbars": true,
    "nativeScrollbarsThin": false,
    "nativeScrollbarsLeft": true,
    "selWinScreenshots": true,
    "updateSidebarTitle": true,
    "markWindow": false,
    "markWindowPreface": "[Sidebery] ",
    "ctxMenuNative": false,
    "ctxMenuRenderInact": true,
    "ctxMenuRenderIcons": true,
    "ctxMenuIgnoreContainers": "",
    "navBarLayout": "horizontal",
    "navBarInline": true,
    "navBarSide": "left",
    "hideAddBtn": false,
    "hideSettingsBtn": false,
    "navBtnCount": true,
    "hideEmptyPanels": true,
    "navActTabsPanelLeftClickAction": "none",
    "navActBookmarksPanelLeftClickAction": "none",
    "navTabsPanelMidClickAction": "none",
    "navBookmarksPanelMidClickAction": "none",
    "navSwitchPanelsWheel": true,
    "groupLayout": "list",
[...]

Issue #665, opened 2022-06-16
[v5] History advanced click support


megamorphg:
## Description
Currently, the history panel only supports opening in a new tab.
Would be great to support the typical advanced clicking options:
- left click to open in current tab
- middle click to open in new tab
- drag and drop support of history (to be able to open new tab in a specific position, send link somewhere, etc.)

mbnuqw:
Resolved in v5.0.0

PR #975, opened 2023-03-05
Support middle-click on history items

jayeheffernan:
This PR adds some basic support for middle-clicking on links in the history panel.

If the user has the setting "Mouse > Bookmarks actions > Middle-click on the bookmark" set to "open in new tab", so that middle-clicking a bookmark opens it in a new tab without navigating away from the bookmarks panel or resetting the search input, then middle-click will behave the same way in the bookmarks panel.

As referenced:

- Here: https://github.com/mbnuqw/sidebery/issues/780#issuecomment-1263307157, and
- Here: https://github.com/mbnuqw/sidebery/issues/665,

...middle-clicking these items currently does nothing.

Ideally the extension would support a range of mouse settings for the history panel, as it does for the bookmarks panel. I've not implemented that here though. Seems to me that a user would rarely want the on-click behavior of history items to be different to that of bookmarks, so the approach here is just to pull from the bookmarks' settings.

### DIFF
--- a/src/services/history.actions.ts
+++ b/src/services/history.actions.ts
@@ -206,7 +206,7 @@ export function scrollToHistoryItem(id: string): void {
   if (el) el.scrollIntoView({ behavior: 'smooth', block: 'nearest' })
 }
 
-export async function openTab(item: HistoryItem): Promise<void> {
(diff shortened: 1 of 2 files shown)


mbnuqw:
Thanks!

Issue #876, opened 2022-12-05
Auto-export every X minutes/hours, or, SAVE/SYNC snapshots to the cloud


wsalopek:
## Description
I'd like to be able to auto-export, (and only snapshots), every X minutes/hours/days, to a user-selectable folder on my hard drive.

And I'd like to be able to IMPORT that snapshots export, to a copy of Sideberry running on another computer.

Thanks...

azizLIGHT:
Would this work as a workaround? 
Make a script to backup the firefox profile directory, and maintain a list of these automated and dated backups to restore into

Would taking a backup of the Firefox profile dir while Firefox is running make incoherent backups of sidebery tree? My firefox is almost always running so live backups are the only way. I wonder if those "live" backups would work to restore sidebery tree without error

domportera:
it would be extra amazing if you could just specify a raw download location (I.e. C:/SomeFolder/SomeOtherFolder) that these auto backups would go to. then any file-syncing cloud storage system would just work out of the box without any additional API dependencies

edit: reviewing the PR, theoretically it should be possible to do this provided your desired backup location is on the drive by abusing relative paths (i.e. `../../../../CloudFolder/SomeOtherCloudFolder`). If the backup location is on a different drive. then a symlink might be required.

Regarding the UX issue discussed in the PR #1047, personally i think most people dont have "Save As" as their default. And if they do, maybe the auto-backup system comes with a warning if using non-relative paths about this focus-stealing annoyance with that FF setting

mbnuqw:
Available from v5.0.0 - as an auto export option (#1047).

PR #1047, opened 2023-05-23
Auto export snapshots

gotjoshua:
work in progress, very interested in feedback

i want to have snapshots show up in my logseq graph automatically... so i hacked together this mvp

@mbnuqw i really dig the architecture... very easy to dive into your code (mostly ; -)


mbnuqw:
Thank you for starting work on it! It's an implementation of #876 (if I'm not mistaken), and I'll merge it when it's done.

A few first thoughts:
- Permission "downloads" should be optional, but since this PR is huge by itself I can do that myself after merge
- Configuration of markdown format (nested list/only headers), probably should be moved to a different PR

gotjoshua:
Cool. Glad you're into it.
Yeah it's a partial solution for #876 
Good call about the mark down indentation style belonging in a separate pr...

I will also need feedback about the refactoring i did... Moving some functions to the actions file that were previously in the Vue file and splitting prepare and export.

I'll also need an explanation of what happens in minify snapshot, cuz it kept me very busy trying to understand what it means for a tab to equal -1

mbnuqw:
> refactoring

Refactoring is ok. Cleaning up of the `snapshots.vue` was planned, but I always procrastinate on such tasks :)


> Meaning of the "-1" values in snapshots

It means that the value is the same as in the previous snapshot (`SnapStoreMode.Unchanged`). Visual example:

| Snapshot1 | Snapshot2 | Snapshot3 | Normalized Snapshot3
|--|--|--|--|
| tab1 | -1 | -1 | tab1 |
| tab2 | tab4 | -1 | tab4 |
| tab3 | -1 | tab5 | tab5 |

Call the `Snapshots.getNormalizedSnapshot()` when you need `NormalizedSnapshot`. I also just found that `Snapshots.getNormalizedSnapshot` incorrectly handle some cases - when the previous snapshot is not found it should return `undefined` and we need to show error message to user (but I'll fix this later).

Also, please, use prettier: `npx prettier -w src/services/snapshots.actions.ts`, `npx prettier -w src/page.setup/components/snapshots.vue`, etc...

gotjoshua:
> Also, please, use prettier: `npx prettier -w src/services/snapshots.actions.ts`, `npx prettier -w src/page.setup/components/snapshots.vue`, etc...

should i take time to get this happening automatically on save via eslint? or you prefer the manual step?

also, `npx prettier -w ./src/ ` also changes some html files that i didn't touch, do you prefer to keep them unpretty? or should i include them in the final PR?

gotjoshua:
> incorrectly handle some cases

ok, maybe i ran into some of those situations, but i anyway managed to work around them by calling auto export before the minification...


and many thanks for the explanations and collaboration!

mbnuqw:
> prettier

I meant to selectively run prettier only for changed files, but never mind, I'll setup a Github action to automate this on PR, so you don't need to run prettier manually. (I'll also update `.prettierignore` with html files).

> and many thanks for the explanations and collaboration!

thank You for spending your time on this

gotjoshua:
> Good call about the mark down indentation style belonging in a separate pr...

i do agree that it might have been better, but it is not so easy for me to revert those changes line by line.

How important is it for you @mbnuqw ?

Do you like the feature idea in general? (i find it really nice for pasting or importing into nested knowledge tools like Logseq)

gotjoshua:
do you have any time for an interim code review?

features are basically working now, and I'll take some more time to clean up and review over the weekend.

src/utils.ts
```diff
@@ -162,10 +162,10 @@ export function uDate(ms: number, delimiter?: string, dayStartTime?: number): st
 /**
  * Get time string from unix seconds
  */
-export function uTime(ms: number, delimiter = ':', sec = true): string {
+export function uTime(ms: number, delimiter = ':', sec = true, min = true): string {
   const dt = new Date(ms)
   let time = `${dt.getHours()}`.padStart(2, '0')
-  time += delimiter + `${dt.getMinutes()}`.padStart(2, '0')
+  if (min) time += delimiter + `${dt.getMinutes()}`.padStart(2, '0')
```
mbnuqw, line 168: Remove `min` argument.

src/types/settings.ts
```diff
@@ -1,6 +1,7 @@
 import { SETTINGS_OPTIONS } from '../defaults'
 
 export interface SettingsState {
+  [x: string]: any
```
mbnuqw: Remove `[x: string]: any`

src/services/snapshots.actions.ts
```diff
@@ -11,13 +11,22 @@ import { Store } from 'src/services/storage'
...
+  SnapshotState,
+} from 'src/types/snapshots'
 import { Containers } from './containers'
 import { DEFAULT_CONTAINER_ID } from 'src/defaults/containers'
 import { PanelType } from 'src/types/sidebar'
 import { ItemInfo } from 'src/types/tabs'
 import { Info } from './info'
 import { createDefaultSidebarConfig, getSidebarConfigFromV4 } from './sidebar-config'
+import { Favicons } from './favicons'
+import { nextTick } from 'vue'
```
mbnuqw: Remove `import { nextTick } from 'vue'` (b/c it's unused, and I'm trying to import view lib only in components and top-lvl scripts like sidebar.ts or setup.ts)

src/services/snapshots.actions.ts
```diff
@@ -29,14 +38,17 @@ let currentSnapshot: Snapshot | undefined
...
   if (!Info.isBg) return await IPC.bg('createSnapshot')
 
   // Get snapshot src data and current snapshots list
   let waiting
   try {
+    console.log('updateBgTabsTreeData')
+    await Tabs.updateBgTabsTreeData(), console.log('getting store')
     waiting = await Promise.allSettled([
       browser.storage.local.get<Stored>(['sidebar', 'containers', 'snapshots']),
-      Tabs.updateBgTabsTreeData(),
```
mbnuqw, line 39: Revert this change (I guess it's a debugging, but just to be sure)

src/services/snapshots.actions.ts
```diff
@@ -136,6 +158,22 @@ export async function createSnapshot(auto = false): Promise<Snapshot | undefined
   return currentSnapshot
 }
 
+export async function exportSnapshot(snapshot: Snapshot) {
+  const prepared = await prepareExport(snapshot)
+  if (!prepared || !browser?.downloads)
+    return console.warn('failed attempt to export snapshot', { snapshot, prepared, browser })
```
mbnuqw: Use internal `Logs.warn('Snapshots.exportSnapshot: Cannot export snapshot')`

src/services/snapshots.actions.ts
```diff
@@ -420,14 +459,14 @@ async function adaptContainers(snapshot: NormalizedSnapshot): Promise<void> {
...
-  if (!stored.sidebar) return
+  if (!stored.sidebar?.nav) return
 
   // Find last index of tabs panel
   let lastStoredTabsPanelIndex = stored.sidebar.nav.length
   while (lastStoredTabsPanelIndex-- > 0) {
     const storedId = stored.sidebar.nav[lastStoredTabsPanelIndex]
     if (storedId === undefined) break
-    const storedPanel = stored.sidebar.panels[storedId]
+    const storedPanel = stored.sidebar?.panels?.[storedId]
```
mbnuqw, line 477: After `if (!stored.sidebar?.nav) return` (which is a good change) it's safe to skip optional chaining `stored.sidebar?.panels?.` because without `stored.sidebar` function will be interrupted and `panels` is a required prop in `SidebarConfig` type.

src/services/snapshots.actions.ts
```diff
@@ -436,7 +475,7 @@ async function adaptTabsPanels(snapshot: NormalizedSnapshot): Promise<void> {
   for (let i = 0; i < snapshot.sidebar.nav.length; i++) {
     const snapId = snapshot.sidebar.nav[i]
     if (snapId === undefined) continue
-    const snapPanel = snapshot.sidebar.panels[snapId]
+    const snapPanel = snapshot.sidebar?.panels?.[snapId]
```
mbnuqw, line 486: It's also safe to not use an optional chaining in `snapshot.sidebar?.panels` because it's a `NormilizedSnapshot` which means `sidebar` prop is required.

src/services/snapshots.actions.ts
```diff
@@ -467,7 +506,7 @@ async function adaptTabsPanels(snapshot: NormalizedSnapshot): Promise<void> {
     }
 
     for (const storedId of stored.sidebar.nav) {
-      const storedPanel = stored.sidebar.panels[storedId]
+      const storedPanel = stored.sidebar?.panels?.[storedId]
```
mbnuqw, line 517: Same

src/services/snapshots.actions.ts
```diff
@@ -619,6 +658,111 @@ export async function removeSnapshot(id: ID): Promise<RemovingSnapshotResult> {
...
+      for (const tab of panel) {
+        const container = tab.containerId ? snapshot.containers[tab.containerId] : undefined
+        // if(tab === -1) {
+        //   console.warn('tab -1', {tab,panel})
+        //   continue
+        // }
+
+        let panelState = panelsById[tab.panelId]
+        if (!panelState) {
+          let panelConfig = snapshot.sidebar?.panels?.[tab.panelId]
```
mbnuqw, line 721: No need for optional chaining (b/c type is `NormilizedSnapshot`)

src/services/snapshots.actions.ts
```diff
@@ -683,7 +827,8 @@ export function convertFromV4(oldSnapshots: Snapshot_v4[]): Snapshot[] {
 
         // Check panel
         if (snapTab.panelId !== -1) {
-          const panelConf = snapshot.sidebar.panels[snapTab.panelId]
+          const panelConf = snapshot.sidebar?.panels?.[snapTab.panelId]
```
mbnuqw, line 838: No need to opt chaining b/c snapshot is `NormilizedSnapshot` and sidebar is `SidebarConfig`

src/services/snapshots.actions.ts
```diff
@@ -736,34 +881,76 @@ export function updateV4GroupUrls(snapshot: NormalizedSnapshot): void {
...
+  return stored.snapshots
+}
+
+export async function prepareExport(snapshot: Snapshot | SnapshotState) {
+  // const parsed = parseSnapshot([snapshot], 0)
+  // console.log({parsed});
+
+  // if(!parsed)return
+  const { id, time, containers, sidebar, tabs } = snapshot //parsed
+  const normSnapshot = { id, time, containers, sidebar, tabs }
```
mbnuqw, line 908: `const normSnapshot` is not `NormalizedSnapshot` and that may cause issues in `convertToMarkdown` fn. It's better to use `NormalizedSnapshot` type instead of `Snapshot | SnapshotState` and update:
- `exportSnapshot(...)` to `exportSnapshot(snapshot: NormalizedSnapshot)`
- in `createSnapshot(...)` fn, pass a `NormalizedSnapshot`- compatible type instead of the `Snapshot` to the `exportSnapshot()`:
```ts
  if (Settings.state.snapAutoExport) {
    void exportSnapshot(currentSnapshot)
  }
```
with
```ts
  if (Settings.state.snapAutoExport) {
    exportSnapshot({
      id: currentSnapshot.id,
      time: currentSnapshot.time,
      containers: stored.containers,
      sidebar: stored.sidebar,
      tabs,
    })
  }
```

src/services/snapshots.actions.ts
```diff
@@ -736,34 +881,76 @@ export function updateV4GroupUrls(snapshot: NormalizedSnapshot): void {
...
+  console.log({ normSnapshot, snapshot })
+
+  const jsonStr = JSON.stringify(normSnapshot)
+  const jsonFile = new Blob([jsonStr], { type: 'application/json' })
+  const mdStr = convertToMarkdown(normSnapshot as NormalizedSnapshot)
+  const mdFile = new Blob([mdStr], { type: 'text/markdown' })
+  return { id, time, containers, sidebar, tabs, jsonFile, mdFile }
+}
+
+export async function getMostRecentSnapshot() {
```
mbnuqw: getMostRecentSnapshot, getMostRecentSnapshotMd are not used, remove?

mbnuqw:
Sorry for delay.

> i do agree that it might have been better, but it is not so easy for me to revert those changes line by line.
> How important is it for you @mbnuqw ?

Not much, leave the features as they are.


> Do you like the feature idea in general?

Yes, and it already was planned (but after v5 release), so it's good to have it implemented earlier.


## Testing:

### Possibly wrong indent in md output with "Tree Friendly MD (add extra bullets for windows and tabs)":
```
# Data/time
  - ## Window 1
  - Tab 1 (pinned)
  - Tab 2 (pinned)
    - ### Panel 1
      - Tab 3
    - ### Panel 2
      - Tab 4
```
It looks like Panel 1 is sub-element of the last pinned tab. What if we use this structure?:
```
# Data/time
- ## Window 1
  - ### Pinned tabs
    - Tab 1 (pinned)
    - Tab 2 (pinned)
  - ### Panel 1
    - Tab 3
  - ### Panel 2
    - Tab 4
```

### Auto-export doesn't work with an empty path ("Path to Export to (relative to downloads)").
It should have the default value ("./" or maybe "./Snapshots" folder).

### Firefox always asks (if configured in FF settings) where to save for auto-created snapshot.
It steals focus and will annoy users. We should use `saveAs: false,` in `browser.downloads.download` for auto-exported snapshots.

### "json" and "both" format options doesn't work

src/services/snapshots.actions.ts
gotjoshua: I don't think i added that, rather just moved it from the old location, I'll try with and without it... And remove if not needed

gotjoshua:
> What if we use this structure

Looks great

Thanks for the detailed review!

Looking forward to get this ready to merge...

src/services/snapshots.actions.ts
mbnuqw: [Initial code:](https://github.com/mbnuqw/sidebery/blob/v5/src/services/snapshots.actions.ts#LL37C1-L40C7)
```ts
    waiting = await Promise.allSettled([
      browser.storage.local.get<Stored>(['sidebar', 'containers', 'snapshots']),
      Tabs.updateBgTabsTreeData(),
    ])
```
[PR:](https://github.com/mbnuqw/sidebery/blob/df72d2d171bf2d1d75b99ad2b6e87cf72981bdf9/src/services/snapshots.actions.ts#LL48C1-L52C7)
```ts
    console.log('updateBgTabsTreeData')
    await Tabs.updateBgTabsTreeData(), console.log('getting store')
    waiting = await Promise.allSettled([
      browser.storage.local.get<Stored>(['sidebar', 'containers', 'snapshots']),
    ])
```

src/services/snapshots.actions.ts
gotjoshua: 👌🙏

gotjoshua:
```md
> - ## Window 1
>   - ### Pinned tabs
>     - Tab 1 (pinned)
>     - Tab 2 (pinned)
>   - ### Panel 1
>     - Tab 3
```

just started on this adjustment and realized that technically the pinned tabs "live" within the panels.
so when i switch panels the pinned tabs switch to the pinned tabs from the other panel

this indicates a structure like this may be better:

```md
- ## Window 1
  - ### Panel 1 Base
    - Pinned
      - Tab 1 (pinned)
      - Tab 2 (pinned)
    - non-pinned Tab 3 ...
 ** OR **
  - ### Panel 2 - Media
    - Tab 1 (pinned)
    - Tab 2 (pinned)
    - non-pinned Tab 3 ...
```

notice that the pinned tabs are different for each panel:
Screenshot from 2023-06-04 12-03-07
Screenshot from 2023-06-04 12-02-35

mbnuqw:
Hm... you're right, ~~it would be better to align markdown rendering with settings (pinned tabs at panel-level or at global-level - "Pinned tabs position" setting).~~

We need to change the way how snapshot is created to distinguish the global pinned tabs (e.g. with tab.panelId === -1) and per-panel (tab.panelId === "PANEL_ID"), because right now it's not possible to get the type of pinned tab. I can do this after this PR will be merged.

mbnuqw:
```md
- ## Window 1
  - ### Panel 1 Base
    - Pinned
      - Tab 1 (pinned)
      - Tab 2 (pinned)
    - non-pinned Tab 3 ...
 ** OR **
  - ### Panel 2 - Media
    - Tab 1 (pinned)
    - Tab 2 (pinned)
    - non-pinned Tab 3 ...
```
Both variants are ok, but the second is better b/c we don't need to translate to other languages "Pinned" (considering that the pinned tabs are marked with 📌).

gotjoshua:
> the second is better

Also semantically feels better to me...

gotjoshua:
> We need to change the way how snapshot is created to distinguish

great that you can do that after merge!

it works now (https://github.com/onezoomin/sidebery/commit/73b62c0e653e506a2a0a326bbbcfc550987ab2eb)

image

but feels rather hackish to remerge the pinned tabs that belong in the panels.

src/services/snapshots.actions.ts
gotjoshua: i thought i would need them, and still might if we want to decouple timing of snapshot creation and export.
i can remove them for now.

gotjoshua: https://github.com/gotjoshua/sidebery/commit/9cd02a6aa9c2e0a40c228914b9e0c106f8d0f4e9

src/services/snapshots.actions.ts
gotjoshua: https://github.com/onezoomin/sidebery/commit/24016a2b19357a8b23b273cd7090b2b45d87e26a

src/services/snapshots.actions.ts
gotjoshua: is this actally better in "parallel" in the allSettled array below? or ok to do it first?

src/utils.ts
gotjoshua: i'd like to keep it for a later option i have in mind (if its harmless enough for you)

gotjoshua:
I ran out of time for this week... 

if you want to finish these up, please feel free... otherwise i'll try to squeeze in some time later next week!

 - [ ] Auto-export doesn't work with an empty path ("Path to Export to (relative to downloads)").
    - It should have the default value ("./" or maybe "./Snapshots" folder).
 - [ ] Firefox always asks (if configured in FF settings) where to save for auto-created snapshot.
    - It steals focus and will annoy users. We should use saveAs: false, in browser.downloads.download for auto-exported snapshots.
 - [ ] "json" and "both" format options doesn't work

mbnuqw:
Alright, thank you for your time. I'll try to finish this and publish a new release soon.

gotjoshua:
> I'll try to finish this

🙏

gotjoshua:
> I'll try to finish this

🙏

PR #1084, opened 2023-06-26
Add "Create new tab" option for single left click on tabs panel

emvaized:
This PR adds new option for single left click on tabs panel — "Create new tab", which allows user to use it instead of double clicking if they want to. When "new tab" button is disabled in settings, it basically allows to use the whole free space on bottom as a big new tab button.


mbnuqw:
Thanks!

PR #1104, opened 2023-07-12
Add support for switching to the nth unpinned tab.

### LINKED ISSUE #976: [v5] Keybindings to scroll.
## Description

With Tree Style Tabs, I've gotten very used to having key bindings for scrolling to the top or bottom of the list of tabs.

I'm not seeing any options for controlling the scrolling of tabs in sidebery, either in steps, or in the full top/bottom.

And it would be quite useful, at least for me.

zelch:
For my personal use, I'll be using the first unpinned tab one, which will always scroll to the top of the tab list, as opposed to jumping to the first pinned tab, which just isn't as useful for me.

This fixes #976.


mbnuqw:
@zelch, maybe add keybindings to scroll to the top/bottom (and by step) of the panel, instead?

zelch:
Hm.

I'm definitely not opposed.  I need to look to see how to do the scrolling, but. :)

mbnuqw:
Sorry if I misunderstood, just to clarify (reading the #976), do you only need to scroll (without tab activation) or do you need tab activation (or both as separate keybindings)?

zelch:
I think that there is value to both.

And I have updated the PR with bindings to scroll to the top and bottom of the active panel.

Page up / Page down on the active panel is definitely something I'd like to add, but I'm definitely going to have to to a lot more digging on how the sidebar scrolling area works first. (And I don't think that should hold up this PR.)

PR #1116, opened 2023-07-17
Add favicon for about:debugging

emvaized:
This PR adds favicon for about:debugging page. It uses a copy of Firefox's built-in `developer` icon: 
[chrome://global/skin/icons/developer.svg](chrome://global/skin/icons/developer.svg)

_Before_: 
image

_After_:


mbnuqw:
Thank you!

Issue #354, opened 2021-03-11
sort name and add folder to container


zpcol:
## Description
if the number of containers that are made is very large, it will be very troublesome if the container names are not in order. it would be neat if there was a "add folder" feature to the container as seen in the bookmarks.

PR #1136, opened 2023-07-26
Add container sorting to bookmark, new tab, reopen, settings menus

fsaresh:
See https://github.com/mbnuqw/sidebery/issues/354 for impetus.

## Context
When opening a new tab, reopening a tab in a container, opening a bookmark in a container, or viewing available containers in Sidebery settings, they currently are displayed in the order of creation rather than alphabetically. 

This changeset introduces a sorting function (`sortContainers`) into the Container Actions file for reusability, and uses that sorting function for those aforementioned context menus and settings menus. I chose to use [`localCompare`](https://www.w3schools.com/jsref/jsref_localecompare.asp) to allow for cross language sorting based on browser settings.

Please note that I haven't worked in TypeScript in a long while and wasn't very familiar with it to begin with, so if there are better ways of doing these things, please let me know :)

## Testing
I followed the steps in the [Contributing Guide](https://github.com/mbnuqw/sidebery/blob/aad7e28c055343e22fdaf6346a4dda5136a271ae/CONTRIBUTING.md#L4) and tested locally, making sure containers were sorted in the new tab, reopen in, open bookmark in, and settings menus.

Before, reopen in menu:
Screen Shot 2023-07-26 at 12 32 45

After, reopen in menu:
Screen Shot 2023-07-26 at 12 36 18

### DIFF
--- a/src/page.setup/components/settings.containers.vue
+++ b/src/page.setup/components/settings.containers.vue
@@ -3,7 +3,7 @@ section(ref="el")
   h2 {{translate('settings.containers_title')}}
   span.header-shadow
   .card(
-    v-for="(container, id) in Containers.reactive.byId"
+    v-for="(container, id) in Containers.sortContainers(Object.values(Containers.reactive.byId))"
     :key="container.id"
     :data-color="container.color")
     .card-body(@click="SetupPage.reactive.selectedContainer = container")
--- a/src/services/containers.actions.ts
+++ b/src/services/containers.actions.ts
@@ -298,3 +298,9 @@ export function getContainerFor(url: string): string | undefined {
 
   return
 }
+
+export function sortContainers(containers: Container[]): Container[] {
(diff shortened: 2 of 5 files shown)


mbnuqw:
LGTM, thank you!

siddhpant:
Any way to have custom ordering or disable sorting?

I'm used to my current setup* and this has thrown me off many times. I use icons, and the container I want to open is now not at the place it used to be.

---

\* I actually put the current ordering deliberately by editing the container settings json file.

siddhpant:
Also, this introduces a mismatch between Firefox's right click context menu "Open Link in New Container Tab" and Sidebery's ordering.

This should be reverted.

mbnuqw:
Ok, I'll make it optional.

siddhpant:
Thank you!

Issue #284, opened 2020-07-28
Use new tabs.warmup() API to improve performance


harshit-j:
## Description
Firefox 79 has a new tabs.warmup() API to improve performance of switching tabs. Can it be implemented in Sidebery?
[https://hacks.mozilla.org/2020/07/firefox-79/](https://hacks.mozilla.org/2020/07/firefox-79/)
> WebExtensions updates
> 
> Starting with Firefox 79, developers of tab management extensions can improve the perceived performance when users switch tabs. The new tabs.warmup() function will prepare the tab to be displayed. Developers can use this function, when they anticipate a tab switch, e.g. when hovering over a button or link.

mbnuqw:
Closing this as implemented

PR #1139, opened 2023-07-28
Implement tabs warmup functionality

emvaized:
After digging in Firefox source code for a while, I found out that it by default [does something](https://searchfox.org/mozilla-central/source/browser/base/content/tabbrowser-tab.js#542) called "tab warmup" — when user hovers the tab, it kind of preloads it's contents, to make the switch feel faster if user decides to click on it. 

Here's the [Firefox Source Docs](https://firefox-source-docs.mozilla.org/browser/base/tabbrowser/async-tab-switcher.html#async-tab-switcher-warming) article with more explanation on this feature. Few words from it:  

> 
> Tab warming allows the browser to proactively render and upload layers to the compositor for tabs that the user is likely to switch to. The simplest example is when a user’s mouse cursor is hovering over a tab. When this occurs, the async tab switcher is told to put that tab into a warming list, and to set its state to STATE_LOADING, even though the user hasn’t yet clicked on it.
> 
> Warming a tab queues up a timer to unload background tabs (if no such timer already exists), which will clear out the warmed tab if the user doesn’t eventually click on it. The unload will occur even if the user continues to hover the tab.

Sidebery doesn't do such warmup, which I checked by setting `browser.tabs.remote.logSwitchTiming` flag in `about:configs` and hovering over Sidebery tabs. And default tabs panel prints various warmup logs in browser console on hovering tabs. 

This PR adds the same functionality in Sidebery using [browser.tabs.warmup](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/warmup) method, and an option to toggle it in settings. Default value is `false` just in case — but probably could be `true`, as native tab panel does it by default.

### DIFF
--- a/src/_locales/dict.setup-page.ts
+++ b/src/_locales/dict.setup-page.ts
@@ -1783,6 +1783,11 @@ Examples: "*", "ctrl+$", "ctrl+alt+g"`,
     de: 'Unten',
     zh_CN: '底部',
     zh_TW: '底部',
+  },  
+  'settings.tab_warmup_on_hover': {
+    en: 'Preload tab contents on mouse hover (tab warmup)',
+    ru: 'Предварительная прогрузка вкладки при наведении мыши',
+    de: 'Tab-Inhalte beim Bewegen der Maus vorladen (Tab-Aufwärmen)',
   },
   'settings.select_active_tab_first': {
     en: 'Select the active tab first',
--- a/src/defaults/settings.ts
+++ b/src/defaults/settings.ts
@@ -82,6 +82,7 @@ export const DEFAULT_SETTINGS: SettingsState = {
   tabsPanelSwitchActMoveAuto: true,
   tabsUrlInTooltip: 'full',
   newTabCtxReopen: false,
+  tabWarmupOnHover: false,
 
   // New tab position
   moveNewTabPin: 'start',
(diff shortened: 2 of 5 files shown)


mbnuqw:
That's actually was in my task list, thank you! I think we don't need settings for it (I'll hide it myself), just make it turned on by default.

mbnuqw:
Issue: #284

emvaized:
> was in my task list

Oh, wow — and I thought I will have to justify including this, as it seemed kind of edgy and unusual :)

emvaized:
I just figured out that it makes no sense to warmup currently active tab on hover (as it's already active). In such case it would be better to warmup the tab which is about  to be activated if user decides to close current tab — or perform the "tab switch" action. However, I'm not sure what method it should use — `Tabs.findSuccessorTab(tab).id` often returns just `1` for me after browser restart. And I guess, it will anyway won't work with `tabFlip()` action, as it uses a different approach of checking Sidebery's history of tabs?

mbnuqw:
Yep, the best we can do here is just `if (!tab.active) browser.tabs.warmup(tab.id)`.

Maybe add warmup on switching tabs with mousewheel, but the time required for implementation is not worth the result (at least in my scenarios, I rarely experience slowdown of tab activation).

PR #754, opened 2022-09-09
Add option to drag-select tabs with middle click to close

emvaized:
Sidebery is currently capable of drag-selecting tabs with right click, e.g. use press the button, move it upwards or downwards to select few more tabs, and then release it to open context menu for all selected tabs.

This commit adds the same functionality for middle click, in order to close multiple tabs at once.
New options sits in `Sidebery settings` > `Mouse` > `Use multi-selection when closing tabs using middle-click`, and is disabled by default.

middle-click-drag-select

PR #1140, opened 2023-07-28
Implement options for Ctrl/Shift middle click on tabs

emvaized:
Default is `discard` for Ctrl + middle click, and `duplicate` for Shift + middle click. Setting option to `none` will revert to the current behavior (serve as a regular middle click)


src/sidebar/components/tab.vue
```diff
@@ -187,6 +187,33 @@ function onMouseDown(e: MouseEvent): void {
   else if (e.button === 1) {
     e.preventDefault()
     Mouse.blockWheel()
+
+    if (e.ctrlKey) {
```
zelch, line 191: This isn't really following the same style as the existing code, where instead of returning, it makes sure that everything is an else if.

That would mean removing the return statements below, and then the existing lines 202 - 204 would become } else if ... {

Same deal on existing lines 215-217.

mbnuqw: Fair point, I'll update it after merge.

mbnuqw: Hm, no, not fair point :), it's simpler to use `return` statement to get the 'fallback' logic (with 'none' values)

emvaized: Yes, fallback to 'none' is exactly why I made it in `return` style. Some people may rely on Ctrl+Middle click to behave just as regular middle click (for whatever reason). 

I was also thinking about writing it in a bit more complex way, so that actions like "duplicate" and "discard" would work with middle click [drag selection](https://github.com/mbnuqw/sidebery/pull/754)— but this way it wouldn't work with "Edit title" action (and I really wanted to have it there), so I guess maybe next time. It's going to be too many checks to handle all possible actions this way.

PR #1141, opened 2023-07-28
Ignore single left click on tabs panel when context menu is open

emvaized:
This PR applies [this patch](https://github.com/mbnuqw/sidebery/commit/9eb70ac5327429f55ecd16e88e36359832175261) for single left click on tabs panel as well. 

It also removes the check for native context menu. The problem, which original patch (as per my understanding) tries to solve, happens to me with native context menu enabled as well — when I have context menu open and perform left click outside in order to close it, the click event gets passed to the Sidebery element under cursor. It happens to me on Windows 11 with "Native context menu" option enabled. And it doesn't happen without the "native menu" check — I think it's not needed anyway, as in the case when left click gets consumed by native context menu's overlay or backdrop (and such context menu gets closed), Sidebery should not receive this event anyway.

### DIFF
--- a/src/sidebar/components/panel.tabs.vue
+++ b/src/sidebar/components/panel.tabs.vue
@@ -80,6 +80,11 @@ function onMouseDown(e: MouseEvent): void {
   if (Selection.isSet()) return
 
   if (e.button === 0) {
+    if (Menu.isOpen) {
+      Menu.close()
+      if (!Settings.state.ctxMenuNative) return
+    }
+
     const la = Settings.state.tabsPanelLeftClickAction
     if (la === 'prev') return Sidebar.switchPanel(-1)
     if (la === 'expand') {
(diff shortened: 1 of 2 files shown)


mbnuqw:
`when I have context menu open and perform left click outside in order to close it, the click event gets passed to the Sidebery element under cursor. It happens to me on Windows 11` Damn, that means that I have to add system-specific logic, because on linux (x11) there is no such event and sidebery keeps thinking that the menu is open which leads to an extra click (3 clicks in general) to activate a tab. Or maybe I'll reset `Menu.isOpen` flag after 1s after opening native context menu (which will also reset selection of target element, but it's ok).

emvaized:
> which will also reset selection of target element

Wouldn't this break functionality when user selected multiple tabs and triggered context menu on them?

mbnuqw:
No, b/c this (resetting selection) will happen after showing context menu. Anyway, it doesn't matter anymore, I [decided to not reset selection](https://github.com/mbnuqw/sidebery/commit/5acc18e111cf237aea19ac8b4daa24f8a7d87d37) since it will be reset after next mouse click or focusOut.

PR #1143, opened 2023-07-29
Warmup successor tab on hovering active tab

emvaized:
I guess it should be better than not doing anything on hovering active tab? This way we lazily store the `successorTabId` in case it's get updated in other places such as `Tabs.removeTabs()` method, and can use this value to warmup the tab that's going to be activated if user closes current tab (either via middle click or close button). 

It seems to me to work better than previously suggested `Tabs.findSuccessorTab(tab)` method call – and it could probably cover even the "flipTab" one when user have a setting to activate previously active tab on close (as it should become the successor tab as well). What do you think?

### DIFF
--- a/src/services/tabs.fg.actions.ts
+++ b/src/services/tabs.fg.actions.ts
@@ -2562,6 +2562,7 @@ function updateSuccession(exclude?: ID[]) {
       browser.tabs.moveInSuccession([activeTab.id], target.id).catch(err => {
         Logs.err('Tabs.updateSuccession: Cannot update succession:', err)
       })
+      activeTab.successorTabId = target.id
       return target
     }
   }
--- a/src/sidebar/components/tab.vue
+++ b/src/sidebar/components/tab.vue
@@ -416,8 +416,15 @@ function onDragStart(e: DragEvent): void {
 }
 
 function onMouseEnter(e: MouseEvent) {
-  if (Settings.state.tabWarmupOnHover && !tab.active) {
-    browser.tabs.warmup(tab.id)
+  if (Settings.state.tabWarmupOnHover) {
+     if (tab.active) {
+        /// warmup successor tab, in case user decides to close active tab
+        const successorTabId = tab.successorTabId;
+        if (successorTabId && successorTabId > -1) browser.tabs.warmup(successorTabId)
+    } else {
+      /// warmup hovered tab
+      browser.tabs.warmup(tab.id);
+    }
   }
 }
 
--- a/src/sidebar/components/tab.vue
+++ b/src/sidebar/components/tab.vue
@@ -417,7 +417,7 @@ function onDragStart(e: DragEvent): void {
 
 function onMouseEnter(e: MouseEvent) {
   if (Settings.state.tabWarmupOnHover) {
-     if (tab.active) {
+    if (tab.active) {
         /// warmup successor tab, in case user decides to close active tab
         const successorTabId = tab.successorTabId;
         if (successorTabId && successorTabId > -1) browser.tabs.warmup(successorTabId)


emvaized:
If you like this idea, probably the same logic could also be implemented on mouse enter in [tabs panel](https://github.com/mbnuqw/sidebery/blob/v5/src/sidebar/components/panel.tabs.vue) — when user has `tabsPanelMiddleClickAction` setting set to close current tab (`rm_act_tab`), it's possible to try to warmup successor tab by checking for active tab's `successorTabId` value

mbnuqw:
Thank you, yes, it's definately better than nothing for the active tab case. There are two uncovered places with `browser.tabs.moveInSuccession` call (I'll fix this after merge).

`cover even the "flipTab" one when user have a setting to activate previously active tab on close` Yep.

`when user has tabsPanelMiddleClickAction setting set to close current tab (rm_act_tab), it's possible to try to warmup successor tab by checking for active tab's successorTabId value`
Hm... maybe we can warmup successorTab of the active tab on every mouseenter over the entire sidebar area? This will automatically cover all future mouse-shortcuts for closing active tab. In this case, we also no longer need to handle mouseenter on the active tab and warmup successorTab inside `tab.vue`.

Another thought: using debounced version of warmup:
```ts
// in tabs.fg.actions.ts
export const warmup = Utils.debounce((id: ID) => {
  if (Tabs.activeId !== id) browser.tabs.warmup(id)
})
```
with ~100ms delay will prevent extra calls of `browser.tabs.warmup` when user moves cursor over the sidebar with lots of tabs.

emvaized:
> maybe we can warmup successorTab of the active tab on every mouseenter over the entire sidebar area?

I'm not a fan of this idea, as it will probably cause many false warmups. For example, if user's mouse enters the sidebar with intention to switch to another tab, it will warmup the successor tab as well – despite user is not likely to switch to it in this situation. And if user hovers 3 or more tabs in succession on his path for the target tab, we turned out to warm up the successor tab for nothing — as it will be pushed out of the warmup stack by browser. 

> mouse-shortcuts for closing active tab

Most users who rely on such shortcuts probably don't hover sidebar too much before performing those :) And it's also important to keep in mind that the tab is indented to be warmed up right before switching it — as it will be unloaded anyway after 2 seconds. It works good with mouse hover events, but probably will never work with keyboard shortcuts (Ctrl+Tab is an exception, as Firefox seems to warm up the highlighted tab in a popup before you release the Ctrl key).

`debounced version of warmup`

While it sounds good on paper, it would probably eliminate the usefullness of warming tabs. From my understanding of [docs](https://firefox-source-docs.mozilla.org/browser/base/tabbrowser/async-tab-switcher.html#async-tab-switcher-warming), it is designed is such way to utilize exactly these 100-200 ms of time we have between user hovered the tab and clicked on it — and few performance optimizations are built-in Firefox: it warms up no more than 3 tabs at one time (tabs newly added to the warmup queue will push out the oldest one), and if user did not switch to tab within 2 seconds it unloads the warmed up tab to free the memory. 

For this moment I saw 3 implementations of tabs warmup in 3 projects: [Firefox itself](https://searchfox.org/mozilla-central/source/browser/base/content/tabbrowser-tab.js#542), [Vertical Tabs Pane](https://github.com/aminomancer/uc.css.js/blob/ddd387cb98dd98fec58940fe74cace584261ad32/JS/verticalTabsPane.uc.js#L1673) userscript, and in [TST](https://github.com/piroor/treestyletab/blob/4b7b95a0dd9280c44e129d9c98575fa6febcaede/webextensions/sidebar/mouse-event-listener.js#L226) — and neither of them appear to be using debouncing to optimize this feature. 

But idk, maybe debouncing with just very small delay, like 50ms or so, would work? Such delay would be small enough to not steal too much time when user deliberately hovered the exact tab they indend to switch, while probably filtering out unwanted warmups when they quickly move the cursor over tabs.

mbnuqw:
> it will probably cause many false warmups

Same applies to checking the `tabsPanelMiddleClickAction` - user may or may not perform that middle click. Checking exact settings and narrow target area for mouse hover is obviously better, but the warming up feature is designed for possible false positives, so maybe we can use simpler code?


> And if user hovers 3 or more tabs in succession on his path for the target tab, we turned out to warm up the successor tab for nothing — as it will be pushed out of the warmup stack by browser.

Not sure if I correctly understand, isn't it exactly how right now works warmup feature? User moves mouse cursor through the list of tabs warming up everything on the way.


> Most users who rely on such shortcuts probably don't hover sidebar too much before performing those :) And it's also important to keep in mind that the tab is indented to be warmed up right before switching it — as it will be unloaded anyway after 2 seconds. It works good with mouse hover events, but probably will never work with keyboard shortcuts (Ctrl+Tab is an exception, as Firefox seems to warm up the highlighted tab in a popup before you release the Ctrl key).

Sorry for incorrect wording, by `mouse-shortcuts` I meant something like 'middle click on the tabs-panel to close active tab', not keyboard shortcut. Of course, keyboard shortcuts are out of scope here.


> ...about debouncing

I tested 100ms, and it was hard for me to navigate mouse cursor to the target tab, click and cause warmup mismatch. For Firefox itself it's useless, since the area of horizontal tabs contains relatively small amount of tabs and its geometry makes it difficult to move cursor through the tabs, so user usually moves cursor from below directly to the target tab. But, for the vertical list it may be useful... may be :), I didn't notice/measure any practical benefits of debouncing (it was just an experiment)

emvaized:
> Checking exact settings and narrow target area

Yep, that was my exact reasoning about it. When we trigger mouse enter on the whole sidebar, user's intention could be anything, and in most cases it will not be closing current tab. From the examples of usage of this feature I concluded that it should be as specific as possible —  and if we're not certain of intention, it probably is better to not warm up any tab at all. 

When user hovers the tab panel itself (I assume this event is only triggered when hovering the empty space on bottom of the panel?), we can with some certainty conclude only 2 intentions: open new tab (which is already warmed up by Firefox), or close current tab (in case of such setting) — there's still uncertainty, but much better than `mouseEnter` event of the whole sidebar. 

But probably it makes sense to not warmup successor tab in such specific case for now, considering the quite likely "open new tab" intention. 

> user usually moves cursor from below directly to the target tab

I agree on this point. But it should also account for users who modified Sidebery to have bigger tabs as hit targets, and can aim the exact tab quite precisely — or just preadjusted the cursor path before entering the sidebar area. As you mentioned, warmup is designed in such way to allow for one tab below/one tab above "mishover" by providing the buffer of 3 tabs — it acts as some sort of debouncing itself, pushing out tabs out of queue by newly added tabs. Adding a delay to that may probably result in cases where user hovered a tab and clicked it with delay less than 100ms — this tab will not get warmed up before the click, making warmup unused. 

By the way, it should be considered that warmup proccess probably takes some time to get finished as well, especially on heavy-loaded pages like Reddit — when I tested, it sometimes took me around 200-300ms to see the "FINISHED" warmup message in the browser console. If debouncing makes sense to be added here, then only with very small delay like 50ms, just to filter out quick mouse movements over the whole list of tabs (which can even be accidental sometimes). 

But it's just my opinion, as there's no precedent of anyone adding debouncing to this feature, at least not that I've seen. Firefox doesn't debounce it even in the [Ctrl+Tab popup](https://searchfox.org/mozilla-central/source/browser/base/content/browser-ctrlTab.js#417), despite user can hold the `Tab` key and jump across various tabs continuously (even though there's usually only 7 of them in the list).

mbnuqw:
I've re-read docs and src (I should do this more often) of the `browser.tabs.warmup` and it seems you're right, debounce is practically useless and extra delay does more harm than good.


> I assume this event is only triggered when hovering the empty space on bottom of the panel?

No, and that's another problem. There is no dynamic-size element below the tabs, on which we can hang the mouseenter listener. 

If we register mouseenter listener on the `.TabsPanel` root element the event will be triggered right after the cursor has crossed the entire `.TabsPanel` element regardless of whether the cursor is hovering over the tabs or the area below.

It's possible to use `mouseover` events of the `.TabsPanel`, check every crossing of the descendant elements and trigger warmup action only if that element is `.tab-space-filler`, `.bottom-space` or `.scrollable`, but it's clumsy, performance non-optimal solution (it's quite a lot of events fired when the cursor moves over the list of tabs).

I propose to stop on the implementation in this PR.

emvaized:
> check every crossing of the descendant elements

Oh, okay then. I just thought it would work the same as with mousedown event on the tabs panel, which gets consumed if something is in between the mouse cursor and tabs panel, for example a tab. Solution you described obviously sounds way too complicated and unnecessary with very small added benefit.

PR #1166, opened 2023-08-12
update translations

52fisher:
All translations updated
图片


mbnuqw:
Thank you!

src/_locales/dict.setup-page.ts
```diff
@@ -2869,27 +2960,49 @@ Examples: "*", "ctrl+$", "ctrl+alt+g"`,
...
+    de: `Hinweis: Pfad ist relativ zum Download-Ordner.
+%Y - Jahr; %M - Monat; %D - Tag; %h - Stunde; %m - Minute; %s - Sekunde`,
+    zh_CN: `注意：路径是相对于下载文件夹的。
+%Y - 年; %M - 月; %D - 日; %h - 小时; %m - 分钟; %s - 秒`,
+    zh_TW: `注意：路徑是相对于下載檔案夾的。
+%Y - 年; %M - 月; %D - 日; %h - 小時; %m - 分鐘; %s - 秒`,
   },
   'settings.snap_export_path_ph': {
     en: 'Sidebery/snapshot-%Y.%M.%D-%h.%m.%s',
+    ru: 'Sidebery/snapshot-%Y.%M.%D-%h.%m.%s',
```
mbnuqw: If there is no translation for the target language, Sidebery will fallback to 'en', so we don't need to duplicate the same value here.

src/_locales/dict.setup-page.ts
```diff
@@ -2899,8 +3012,8 @@ Examples: "*", "ctrl+$", "ctrl+alt+g"`,
       return 'минут'
     },
     de: n => (n === 1 ? 'Minute' : 'Minuten'),
-    zh_CN: n => (n === 1 ? 'minute' : '分'),
-    zh_TW: n => (n === 1 ? 'minute' : '分'),
+    zh_CN: '分',
+    zh_TW: '分',
```
mbnuqw: In this case, we can use just `zh: '分'`, which will cover both `zh_CN` and `zh_TW`

52fisher: Now I have improved some Chinese translations

52fisher: Now I have improved all the translations, including the previous duplicated the same value.

mbnuqw:
Thanks again!

Issue #1200, opened 2023-08-28
[v5] Need blacklist for detect externally opened tab.


zelch:
## Description

It turns out that the GlobalProtect VPN arrangement used by my company breaks in interesting ways if Sideberry reopens the externally opened tab in another container.  I suspect that it's the reopening that does it.

Regardless, this means that I need to be able to blacklist specific hosts from being reopened.

zelch:
So, I was somewhat wrong about what was breaking, and I finally figured it out.

I opened a PR to fix the actual bug. :)

PR #1205, opened 2023-08-31
Don't reopen during POSTs.

zelch:
This can happen, for example, when something loads a file:// URL, that then has static form data and javascript to immediately submit it.

We then turn the POST with data in the default container into a GET with no data in the target container.

That doesn't work very well.

As such, let's just refuse to reopen a tab in a new container during a POST, we can't bring the data with us (that I've figured out how to do anyhow).

This was the actual problem that had me originally open #1200 .

### DIFF
--- a/src/services/web-req.fg.ts
+++ b/src/services/web-req.fg.ts
@@ -24,6 +24,10 @@ function onBeforeRequestHandler(info: browser.webRequest.ReqDetails): optBlockin
       const dst = { panelId: tab.panelId, containerId: tab.reopenInContainer }
       const item = { id: tab.id, url: info.url, active: tab.active, index: tab.index }
       delete tab.reopenInContainer
+      if (info.method == "POST") {
+        Logs.warn("onBeforeRequestHandler not reopening, POST.")
+        return {}
+      }
       Tabs.reopen([item], dst)
       return { cancel: true }
     }


mbnuqw:
Thank you!

mbnuqw:
> We then turn the POST with data in the default container into a GET with no data in the target container.

I think, in this case it would be better to just handle only GET method (to avoid possible issues with other non-GET requests): https://github.com/mbnuqw/sidebery/commit/dd44210079ddb27256f0b3e30d51ec053cbbc080

zelch:
> I think, in this case it would be better to just handle only GET method (to avoid possible issues with other non-GET requests): https://github.com/mbnuqw/sidebery/commit/dd44210079ddb27256f0b3e30d51ec053cbbc080

I agree, however I'm going to strongly suggest that we still 
```
delete tab.reopenInContainer
```

Otherwise, I believe, we may end up reopening in a new container on the next GET in the window.

Which doesn't sound bad, until the site that you're POSTing to sets cookies based on the POST data.

mbnuqw:
> I agree, however I'm going to strongly suggest that we still...

Oops, forgot about that. Thanks for noticing this.

zelch:
No worries, I had a few...  Missteps while figuring out what was going on. :)

PR #1206, opened 2023-09-01
Better shortcuts

zelch:
Currently, the mozilla documentation states that for secondary modifiers:
    
> If supplied, this must be either "Shift" or (for Firefox ≥ 63) any one of "Ctrl", "Alt", "Command", or "MacCtrl". Must not be the modifier already used as the main modifier.
    
This commit adds Ctrl, Command, and MacCtrl to the list of allowed secondary modifiers, allowing stuff like MacCtrl+Ctrl+Period as a binding.
    
Of course, that renders quite confusingly (Ctrl+Ctrl+Period), so I did a second change to render MacCtrl as ⌃, which is UTF 2302 Up Arrowhead.  This appears to be what MacOS is using to render it.

At the very least, it's the symbol they use on their website for it. :)

This gives us ⌃+Ctrl+Period, which is at least less confusing, even if I wish that Command came through as Command.

### DIFF
--- a/src/page.setup/components/keybindings.keybinding.vue
+++ b/src/page.setup/components/keybindings.keybinding.vue
@@ -53,7 +53,7 @@ function changeKeybinding(cmd: Command): void {
 
 function normalizeShortcut(s?: string): string {
   if (!s) return '---'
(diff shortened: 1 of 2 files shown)


mbnuqw:
Great, thank you!

PR #1231, opened 2023-09-17
Fix search with about:blank tabs.

zelch:
Currently, search fails if a panel has any tabs which are open to about:blank.

This check fixes the issue, allowing search to work.


mbnuqw:
Thanks

PR #1276, opened 2023-09-25
feat: Added a "Toggle branch" keyboard shortcut setter

### LINKED ISSUE #1211: Dual Function Shortcut: Toggle between folded/expanded states of branch (Merge folding & expanding)
## Description

Providing two shortcuts, one for expanding and one for folding, is unnecessary based upon the binary nature of the state of a branch. Therefore, lets merge those two shortcuts into one single action and allow users to only need activate one keybind whilst retaining the ability to execute both actions.

paul-ohl:
This comes to resolve [Issue #1211](https://github.com/mbnuqw/sidebery/issues/1211).

I added a "Toggle selected branch" keyboard shortcut.
The shortcut uses already existing functions, I added very little logic.

The option lacks translation, I have no idea how these are usually done.
I can add them but I will be using google translate so they might not be very good.

I invite you to be extra careful with my PR, this is my first time contributing code to a project.

### DIFF
--- a/src/_locales/dict.browser.json
+++ b/src/_locales/dict.browser.json
@@ -214,6 +214,9 @@
     "zh_CN": "展开选定的分支",
     "zh_TW": "展開選定的分支"
   },
+  "KbToggleBranch": {
+    "en": "Toggle selected branch"
+  },
   "KbFoldInactBranches": {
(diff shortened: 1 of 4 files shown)


mbnuqw:
Thank you!

PR #1298, opened 2023-09-29
Improve Traditional Chinese translation.

llc0930:
Mainly updated the Traditional Chinese translation.
Simplified Chinese only corrects obvious errors and does not unify the translation style, because I dare not say that I am familiar with it.

Some unused translation items are marked `TODO:` for you.
The positions of some translation items have been moved to facilitate translation.
But both are not quite complete yet. I'm a little busy recently...

At least there should be no problem with the Traditional Chinese part. I hope there will be a chance to use a better quality translation next time a new version is released.
Thanks for your work, this is a useful Add-ons.


mbnuqw:
Thank you, could you please remove changes that are not related with "Traditional Chinese translation" like:
- A separator in default menu options list (`src/defaults/menu.ts`)
- Notes in `src/page.setup/components/keybindings.vue`
- And the fix for incorrect conditional in `src/page.setup/components/settings.mouse.vue`

You can create separate PRs for these.

llc0930:
Okay, I separated them out, please pay attention to the order when pulling.
By the way, [Firefox updated their account names](https://support.mozilla.org/en-US/kb/firefox-accounts-renamed-mozilla-accounts).
For the translation of `settings.sync_notes`, I only fixed the Traditional Chinese part because I'm not sure about the renaming status in other languages.

llc0930:
After more than a month I finally got my account back... Updated translation for the latest settings.

mbnuqw:
Thanks!

PR #1326, opened 2023-10-21
Confirm popup by pressing the enter key

into-the-v0id:
Currently popups can be closed by pressing the Esc Key. However, they cannot be confirmed by pressing the Enter Key. This PR adds this functionality.


mbnuqw:
Thanks!

PR #1338, opened 2023-10-29
Optimize moving tabs inside the same window

Lej77:
I noticed that moving all tabs inside a window to a new panel could be quite slow so I took a look at the code in order to optimize it. I suspect that it was the left over `TODO: Do not call this fn if: all tabs go one after another and srcIndex === dstIndex` that was the cause of my issue so I implemented a check for that condition and when I ran the code in the debugger the `browser.tabs.move` function was indeed not called when moving all tabs to a new panel.

### DIFF
--- a/src/services/tabs.fg.move.ts
+++ b/src/services/tabs.fg.move.ts
@@ -160,6 +160,10 @@ export async function move(
     return
   }
 
+  // Info about the moved tabs previous state:
+  const oneAfterAnother = tabs.every((tab, ix) => ix === 0 || tab.index === tabs[ix - 1].index + 1)
+  const srcIndex = tabs[0].index
+
   const ids = []
   let dstIndexIncluded = -1
   let prevIndex = 0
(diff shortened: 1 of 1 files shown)


mbnuqw:
Thank you.

`canSkipMove` gets the `true` value when you drag and drop the tab to one position down/up, so native tabs are not updated. This happens because the local state of tabs has been updated at this (line 290) moment, and the `tabs[0].index` has its final (as after the move) value.

We need to gather all needed info before the change of the local state of tabs:
```ts
const oneAfterAnother = tabs.every((tab, ix) => ix === 0 || tab.index === tabs[ix - 1].index + 1)
const srcIndex = tabs[0].index
// ^ Somewhere before line ~166
```

And check if we `canSkipMove` using this info (instead of [290..294](https://github.com/mbnuqw/sidebery/pull/1338/files#diff-6c83957cfa1d034489f9eb648ca3b44f266c9213dcac36cff3c0e6919ef888e7R290-R294)):
```ts
const samePosition = srcIndex === tabs[0].index
const canSkipMove = oneAfterAnother && samePosition
```

Lej77:
My bad, should have thought about the internal state updating. Anyway, I made the suggested changes and I have checked that the updated code does actually move the tabs in the native tab bar. So I think it should work now!

Issue #1308, opened 2023-10-04
Search pinned tabs


morphles:
## Steps to reproduce
- Open Firefox in [new profile](https://support.mozilla.org/en-US/kb/profile-manager-create-remove-switch-firefox-profiles)
- Install Sidebery
- Bind search in Sidbery settings
- use search to look for tabs

## Expected behavior
ability to navigate to pinned tabs with names/titles matching searched term, just like for not pinned tabs

## Actual behavior
no access to pinned tabs, no highlight of pinned tabs even if name/title matches

## Misc
Maybe I'm doing something wrong? :)

PR #1347, opened 2023-11-04
Search improvements

zotanmew:
This PR fixes searching for pinned tabs, resolving #1308

It also adds the option to search through all panels instead of just the current one.

### DIFF
--- a/src/_locales/dict.setup-page.ts
+++ b/src/_locales/dict.setup-page.ts
@@ -1476,6 +1476,14 @@ Beispiele: "*", "ctrl+$", "ctrl+alt+g"`,
     zh_CN: '切换到历史记录',
     zh_TW: '切換到歷史紀錄',
   },
+  'settings.search.pinned_tabs_first': {
+    en: 'Return pinned tabs first',
+    de: 'Angeheftete Tabs zuerst anzeigen',
+  },
+  'settings.search.all_panels_always': {
+    en: 'Always search all panels',
+    de: 'Immer alle Panele durchsuchen',
+  },
 
   // - Tabs
   'settings.tabs_title': {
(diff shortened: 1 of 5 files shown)


src/_locales/dict.setup-page.ts
```diff
@@ -1476,6 +1476,14 @@ Beispiele: "*", "ctrl+$", "ctrl+alt+g"`,
     zh_CN: '切换到历史记录',
     zh_TW: '切換到歷史紀錄',
   },
+  'settings.search.pinned_tabs_first': {
+    en: 'Return pinned tabs first',
+    de: 'Angeheftete Tabs zuerst anzeigen',
+  },
```
llc0930: `zh_CN: '首先显示已固定标签页',`
`zh_TW: '首先顯示已釘選分頁',`

src/_locales/dict.setup-page.ts
```diff
@@ -1476,6 +1476,14 @@ Beispiele: "*", "ctrl+$", "ctrl+alt+g"`,
...
+  'settings.search.all_panels_always': {
+    en: 'Always search all panels',
```
llc0930: `en: 'Search in all panels',`

src/_locales/dict.setup-page.ts
```diff
@@ -1476,6 +1476,14 @@ Beispiele: "*", "ctrl+$", "ctrl+alt+g"`,
...
+    de: 'Immer alle Panele durchsuchen',
+  },
```
llc0930, line 1490: `zh_CN: '在所有面板中搜索',`
`zh_TW: '在所有面板中搜尋',`

llc0930 (commented):
Improve translations for you.

mbnuqw:
Thanks for working on this, but I can't merge it because of the lack of visual distinction between different types/sources of tabs in result list. In other words, it's impossible to understand which tab is pinned and which is not, or which tab belongs to the active panel and which belongs to another panel. Second reason - rendering tabs of one panel in another panel is conceptually wrong.

To bring the actual search (instead of current filtering) with support of pinned tabs, all panels and all windows, I plan to render results in a separate list (maybe popup or special panel-like layout) with sections for results: Window/Panel/Pinned tabs/Normal tabs...

Regarding the #1308 issue. To fix it, we need to add tab filtration for PinnedTabsBar component and update behavior of selection in "search" mode.

PR #1390, opened 2023-12-15
Fixed: setting scroll area note

King-of-Infinite-Space:
The note for setting tab scroll area width is backwards, according to the implementation
https://github.com/mbnuqw/sidebery/blob/2cbf82360a96ff6014a5f847f807f1542470c1da/src/services/sidebar.actions.ts#L175-L181


llc0930:
You have to be kidding... Actually, by setting this value to 0, you'll find that it is used to reserve scroll width. The only thing wrong is that the translation of zh_TW is reversed. This error has been fixed in the PR I submitted.
https://github.com/mbnuqw/sidebery/blob/5b0e60019ab61f8ec86568c385a4becf1e40b440/src/_locales/dict.setup-page.ts#L3139-L3152

mbnuqw:
Hm... it works as intended: Scroll area (rotating mouse wheel will scroll the content) is on the right side and with the correct width when value is positive and vice versa.

King-of-Infinite-Space:
I'm not getting the intended behavior (v5.0.0, firefox 121.0.1). With a positive value, I can scroll if cursor is at position 1 but not at position 2.

Also, if I set it to smaller positive value, I actually get larger scroll area, so the width calculation seems incorrect, please see this issue #1391

mbnuqw:
It's probably a misunderstanding of this setting, and I need to fix option's label to clarify its meaning. Or I just can't reproduce the issue.

With the positive value - You scroll content when the cursor over the right area 2 (at given width) and switch between tabs when cursor over the left area 1 (remaining width).
With the negative value - You scroll content when the cursor over the left area 1 (at given width (abs)) and switch between tabs when cursor over the right area 2 (remaining width).

King-of-Infinite-Space:
Thanks for clarifying. I see my misunderstanding. The setting is under "Switch tabs with scroll wheel" so I assumed the width is for switching tabs. It will be helpful to clearly label what the area is for. I'll close #1391.

PR #1424, opened 2024-01-21
Feat: Added one panel switch per scroll setting

ImTheSquid:
Added an option in settings to only allow for one panel switch when using the horizontal scroll to switch feature, similar to how Arc works. I added the base English translation for the text.

### DIFF
--- a/src/_locales/dict.setup-page.ts
+++ b/src/_locales/dict.setup-page.ts
@@ -3009,6 +3009,9 @@ Beispiele: "*", "ctrl+$", "ctrl+alt+g"`,
     zh_CN: '使用滚轮水平滚动',
     zh_TW: '使用滾輪水平滾動',
   },
+  'settings.one_panel_switch_per_scroll': {
+    en: 'Limit to one panel switch per scroll',
+  },
   'settings.h_scroll_action_switch_panels': {
     en: 'switch panels',
     ru: 'переключения панелей',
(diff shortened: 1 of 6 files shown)


mbnuqw:
Thanks. I think it would be better if we moved your debouncing algorithm to [Sidebar.switchPanel](https://github.com/mbnuqw/sidebery/blob/af2dc98632cedb9ea24542bd5bbf3a57b8cef509/src/services/sidebar.actions.ts#L1243) to avoid double debouncing. Plus, there is already a setting for configuring specific delay time `navSwitchPanelsDelay` (50ms on my hardware was not enough).

Something like that (in sidebar.actions.ts):
```ts
function startPanelSwitchingDebounceTimer(delay: number) {
  switchPanelPause = setTimeout(() => {
    clearTimeout(switchPanelPause)
    switchPanelPause = undefined
  }, delay)
}

let switchPanelPause: number | undefined
export function switchPanel(
  dir: 1 | -1,
  ignoreHidden?: boolean,
  withoutTabCreation?: boolean,
  restartDebouncer?: boolean
): void {
  // Debounce switching
  const delay = Settings.state.navSwitchPanelsDelay ?? 128
  if (switchPanelPause) {
    if (restartDebouncer) { // + check switched scroll direction
      clearTimeout(switchPanelPause)
      startPanelSwitchingDebounceTimer(delay)
    }
    return
  }
  if (delay > 0) startPanelSwitchingDebounceTimer(delay)

//...
```

And that (in sidebar.vue):
```ts
  if (e.deltaX > 0) return Sidebar.switchPanel(1, true, false, Settings.state.onePanelSwitchPerScroll)
  if (e.deltaX < 0) return Sidebar.switchPanel(-1, true, false, Settings.state.onePanelSwitchPerScroll)
```

ImTheSquid:
Makes sense, I'll switch it over.

PR #1451, opened 2024-02-07
Add GHA build

gliptak:
https://github.com/gliptak/sidebery/actions/runs/7809214385

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Use Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@v4
       with:
-        node-version: lts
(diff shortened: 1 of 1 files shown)


mbnuqw:
Thanks, but I don't see the need for this Workflow. I plan to make an automated weekly build of the "github" version (incrementing the fourth number of the version: vX.X.X.N+1), in which, instead of building an unsigned archive (like here), it will create a signed `.xpi` (via `web-ext sign`) and update `updates.json`. This way, users of "github" builds will automatically receive the latest updates. Most likely, I will do this when preparing the next AMO version.

gliptak:
@mbnuqw the benefit of this setup that contributions will be built/tested and you might even attach an `xpi` specific to the PR

Issue #301, opened 2020-08-18
[Featur Request] Show tab's preview on hover


emvaized:
## Description
As it can be seen, Sidebery already can store tab's screenshots, to display them in tab group (list mode).

So it will be awesome to show a small popup on tab's hover containing preview of tab's content, as well as title and favicon.
I think of a great example of this feature in Vivaldi browser:
![](https://user-images.githubusercontent.com/37851576/90471511-26412880-e127-11ea-80e5-099f164dbe38.png)

emvaized:
@mbnuqw Would it be possible to implement this feature some day? 
If it is problematic to show tooltip with image outside of the sidebar bounds, maybe [Visual Tabs](https://addons.mozilla.org/firefox/addon/visualtab/) approach is possible? 

Thanks

schmurtzm:
"Preview in popup inside sidebar" seems more reactive than "Preview in window" and no flickering. So I prefer the second option ("Preview in popup inside sidebar").
In all cases this is a very nice feature !

emvaized:
@mbnuqw 
Thanks a lot for your detailed answer. I would agree that 2nd approach seems to be the most intuitive and easy to get used to. 

However, I wanted to point out that by 3 (Visual Tabs-like approach) I actually meant that previews should be revealed on hover, not always visible on all tabs — extension has an option for this:

https://user-images.githubusercontent.com/37851576/175440304-d8636f41-b60c-415c-9ad5-19b9ec352280.mp4

This way discarded tabs, which don't provide any preview, simply will not expand on mouse hover.

guigirl42:
Looking forward to this _so very much_. It's a fantastic Vivaldi feature, sorely missed in FF, so if `Sidebery` would be able to provide this, i'll be delighted.

jathek:
Option 1 would be best, as it doesn't affect the sidebar's usability. The flickering is annoying though.

Next best is option 2 imo, since adding the preview to the tooltip isn't too bad when it comes to usability.

Option 3, where tabs change size when you hover over them, seems like it would have a lot of usability issues. Every time you hover over a tab, it looks like you triple the distance to the next tab.

emvaized:
> * Due to https://bugzilla.mozilla.org/show_bug.cgi?id=1271047 it's not possible set initial position of the window, only update it after creation

By the way, it seems that this bug has been resolved 2 months ago...

guigirl42:
> this bug has been resolved 2 months ago

[...]

PR #1463, opened 2024-02-17
Fixed an issue when "Sidebar Side" is set to "right" value.

llc0930:
Do updSize first, then updPosition to ensure that the size of the preview window does not exceed the screen. (#301)


llc0930:
Now except for the flickering of the window, it can basically be used normally.

PR #1537, opened 2024-03-24
fix: L10n zh_TW correction

llc0930:
- Reconfirm warning behavior when closing multiple tabs and how child tabs are handled when closing parent tabs, and correct related translations.
- Other recent translation fixes.


mbnuqw:
Thanks

PR #1608, opened 2024-04-18
chore: manifest v3 (#916)

### LINKED ISSUE #916: Manifest V3 migration
## Links

[WebExtensions Community Group](https://github.com/w3c/webextensions)
[Manifest V3 migration guide](https://extensionworkshop.com/documentation/develop/manifest-v3-migration-guide/)

sayalovesu:
i'm not entirely sure i did everything, this seems like too few changes for a v3 migration. sorry for being dumb in advance. 

one thing im most unsure about is this part:
> The [Scripting API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting) takes over the features of tabs.insertCSS(), tabs.removeCSS(), and tabs.executeScript() and adds capabilities to register, update, and unregister content scripts at runtime.

i've seen quite a few usages of executeScript, but the extension seems to be working fine despite using them? 
and i also didn't really understand what those calls are used for in the first place

also im not sure how do i do a regression test, there don't seem to be any kind of tests in the repo

closes #916

### DIFF
--- a/src/bg/background.ts
+++ b/src/bg/background.ts
@@ -129,7 +129,7 @@ void (async function main() {
 function initToolbarButton(): void {
   Menu.createSettingsMenu()
 
-  browser.browserAction.onClicked.addListener((_, info): void => {
+  browser.action.onClicked.addListener((_, info): void => {
     if (info && info.button === 1) browser.runtime.openOptionsPage()
     else browser.sidebarAction.toggle()
   })
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -28,6 +28,7 @@
     "storage",
     "unlimitedStorage",
     "sessions",
+    "scripting",
     "menus",
     "menus.overrideContext",
     "search",
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -1,5 +1,5 @@
 {
-  "manifest_version": 2,
+  "manifest_version": 3,
   "browser_specific_settings": {
     "gecko": {
       "id": "{3c078156-979c-498b-8990-85f7987dd929}",
@@ -34,7 +34,6 @@
     "theme"
   ],
   "optional_permissions": [
-    "<all_urls>",
     "proxy",
     "webRequest",
     "webRequestBlocking",
@@ -43,6 +42,9 @@
     "clipboardWrite",
     "history",
     "downloads"
+  ],
+  "host_permissions": [
+    "<all_urls>"
   ],
   "sidebar_action": {
     "default_icon": "./assets/logo-native.svg",
@@ -424,7 +426,7 @@
       "description": "__MSG_KbSortPanelTabsByTimeDes__"
     }
   },
-  "browser_action": {
+  "action": {
     "default_icon": "./assets/logo-native.svg",
     "default_title": "__MSG_ActionTitle__",
     "default_area": "navbar",
--- a/src/services/menu.actions.ts
+++ b/src/services/menu.actions.ts
@@ -107,15 +107,23 @@ export function createSettingsMenu(): void {
     id: 'open_settings',
     title: translate('menu.browserAction.open_settings'),
     icons: { '16': 'assets/logo-native.svg' },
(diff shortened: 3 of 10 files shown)


sayalovesu:
the reason im doing this pr in the first place is that this migration is kinda a blocker for chrome support: promise-based apis are only available in mv3, and ext store doesn't allow publishing mv2 anymore

i really wanted to do that, since chrome finally has an api for side panels and it would be really cool to get sidebery (at least partially) working there

mbnuqw:
Thanks for starting working on it!

> i've seen quite a few usages of executeScript, but the extension seems to be working fine despite using them?

`tabs.executeScript` used:
- In [Group page](https://github.com/mbnuqw/sidebery/blob/aa476a47f3663230ee5722d1a3cebffe4faf3192/src/services/tabs.bg.actions.ts#L572) and [Url-placeholder page](https://github.com/mbnuqw/sidebery/blob/aa476a47f3663230ee5722d1a3cebffe4faf3192/src/services/tabs.bg.actions.ts#L515) as a workaround to make these pages persistent between addon update.
- When [unloading tabs](https://github.com/mbnuqw/sidebery/blob/aa476a47f3663230ee5722d1a3cebffe4faf3192/src/services/tabs.fg.actions.ts#L1166) to unload tabs that try to prevent their closing.
- To [pause](https://github.com/mbnuqw/sidebery/blob/aa476a47f3663230ee5722d1a3cebffe4faf3192/src/services/tabs.fg.media.ts#L88)/play media content
- To [show tab preview](https://github.com/mbnuqw/sidebery/blob/aa476a47f3663230ee5722d1a3cebffe4faf3192/src/services/tabs.preview.ts#L146) in `in-page` mode

so all these features will be broken without replacing `tabs.executeScript` with `scripting.executeScript`

> also im not sure how do i do a regression test, there don't seem to be any kind of tests in the repo

There are no tests, since this project for fun, and writing tests is not fun for me... so only manual e2e testing

---

To get it merged, we need:
- Sidebar Styles: Fix box-sizing (mv2 default was: border-box)
- Sidebar Styles: Set default cursor to "default"
- Optimize background to non-persistant mode:
  - [Make sure that event listeners will restart background script if needed](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Background_scripts#move_event_listeners)
  - Make sure that auto-reopening in container or container proxy features are working even if sidebar is closed (hm... somehow prevent bg suspension?)
- Use `scripting.executeScript()` instead of `tabs.executeScript()`
  - Since `scripting.executeScript()` doesn't support `code` prop that is used in Sidebery to provide initial data for Group page, Url-placeholder page or Tab previews, we need to send that data via [messaging system](https://github.com/mbnuqw/sidebery/blob/aa476a47f3663230ee5722d1a3cebffe4faf3192/src/services/ipc.ts).

sayalovesu:
thanks for the details, i appreciate that, they really helped!~

i fixed everything except the "Optimize background to non-persistant mode" part.
it *seems* to be working fine, as in - the background script just restarts as soon as it's terminated:

https://github.com/mbnuqw/sidebery/assets/163348146/fd7204cf-2ea1-48e5-a271-04a955c35c04

definitely not the best way, but considering how much sidebery relies on the background script - i feel like it's good enough

noguxx:
@mbnuqw and @kunokareal
Could you please change the [browser_specific_settings.gecko.id](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_specific_settings#firefox_gecko_properties) to a **string formatted like an email address** instead of a GUID to solve #1629 
My suggestion would be `sidebery@mbnuqw`

sayalovesu:
> Could you please change the... 

im not an expert on extensions, but i think changing that  id might lead to unintended side effects like resetting storage and braking ota updates. 

in any case, i think its better to keep things focused and for you to open a pr of your own that deals with that.

devsk:
Is this still being worked on or is it abandoned?

PR #1698, opened 2024-06-26
feat: open window (from drag) under mouse

rebelonion:
This PR makes new windows opened by dragging a tab from sidebery to outside the container open directly under the mouse.

The new window will open with an offset relative to where the user's mouse was inside the original window, mimicking Firefox's default tab behavior.

Tested on 1080p, 1440p, and multi-window.

<details>
  <summary>Before</summary>

  [Before](https://github.com/mbnuqw/sidebery/assets/87634197/94351ac8-da82-48c7-9120-d874b7ba7dcb)

</details>

<details>
  <summary>After</summary>

  [After](https://github.com/mbnuqw/sidebery/assets/87634197/ce5b19e0-6bed-41a2-a5cb-975c3630fad2)

</details>

### DIFF
--- a/src/services/drag-and-drop.actions.ts
+++ b/src/services/drag-and-drop.actions.ts
@@ -1191,6 +1191,13 @@ export function resetOther(): void {
 
 let dragEndedRecentlyTimeout: number | undefined
 
+let relativeMouseX = 0
+let relativeMouseY = 0
+export function onDragStart(e: DragEvent): void {
+  relativeMouseX = e.clientX
+  relativeMouseY = e.clientY + (window.outerHeight - window.innerHeight) //account for window height
+}
+
 export async function onDragEnd(e: DragEvent): Promise<void> {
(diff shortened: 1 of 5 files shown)


rebelonion:
since ``maximized`` may not be combined with ``left``, ``top``, ``width``, or ``height``, to get a new window to open maximized on the correct screen/monitor, it needs to be opened with the ``normal`` state with the correct ``left`` and ``top``, then maximized.

An odd behavior I found was if the initial window was maximized, the new window would always open maximized first before shrinking, even when given a ``normal`` state. This would ruin the new x and y for the new screen. I got around this by forcing a ``width`` and ``height`` of the new window before maximizing it.
Idk why Firefox always maximizes first. Almost seems like a bug on their side.

rebelonion:
I no longer use this extension, so if someone else wants to use this as a starting point, go ahead.

PR #1813, opened 2024-09-26
Fix webext.run for zen-browser

shanto:
This build script fails to create the profile directory when full path to browser is provided as the last argument. Fixes in this commit solves that, and also adds support for zen browser.


mbnuqw:
Thanks

Issue #686, opened 2022-07-07
Keyboard to [switch to next panel] should cycle and be global


Fred-Vatin:
Firefox | Extension
------- | ---- 
102.0   | 4.10.1

## Description

I am a very keyboard user and I’ve bound `ALT+W` to `switch to next panel`. I find it surprising this shortcut only worked when the sidebar is open and that it doesn’t cycle. I’m coming from [Panorama Tab Groups](https://github.com/projectdelphai/panorama-tab-groups) that works this way.

seppulcro:
Hi! It's never too late to come back, this feature is part of #1829 .
I liked sidebery so much and it fit my workflow so well that I decided to give it a little extra shine.

PR #1829, opened 2024-10-10
feat: Add the ability to cycle through panels (hidden configurable)

seppulcro:
Hi, I've added the ability to cycle through panels infinitely, with an option do go through hidden panels as well.

Solves #686 

[output.webm](https://github.com/user-attachments/assets/376afca2-69b8-41ce-a257-dfd0d1d3f3bd)


ChilledMoth:
Nice, but why are you taking over the default hotkey for 'Open a new tab after the active tab'?

Just leave it undefined or set a new one, idk, 'Alt+M' or something (since 'Alt-Comma' and 'Alt+Period' are the defaults for 'Switch to previous panel' and 'Switch to next panel', respectively).

seppulcro:
> Nice, but why are you taking over the default hotkey for 'Open a new tab after the active tab'?
> 
> Just leave it undefined or set a new one, idk, 'Alt+M' or something (since 'Alt-Comma' and 'Alt+Period' are the defaults for 'Switch to previous panel' and 'Switch to next panel', respectively).

Whoops sorry I've been using that since I've setup it up and must've forgot to change it back.

I've also noticed a typo that I've fixed, and since I started using it for myself yesterday the ability to go only forwards didn't seem enough... so I added the ability to go backwards too in the latest commit.

seppulcro:
@ChilledMoth please re-validate when you have the time so I know if it's up to spec now 😁

llc0930:
I personally think this feature should be added as an optional setting rather than replacing the original setting.

mbnuqw:
Thanks

PR #1835, opened 2024-10-13
Stop changing window title on sidebar disconnect if markWindow setting is off

jackordman:
I am using another extension https://github.com/irvinm/Toggle-Native-Tab-Bar which uses a similar technique to hide native tab bar by adding/removing a " " character from the window title. Closing the sidebar would make the native tab bar reappear even though the Sidebery setting to preface the window title was OFF. This change simply checks that setting before setting the window title to '' on a sidebar disconnect, just like it does on a sidebar connection. Unsure if this will affect anything downstream however.


mbnuqw:
Thanks

Issue #1320, opened 2023-10-13
[v5] Option to Unload tabs by middle clicking Favicons


Crunchbits:
## Description
I find unloading to be very useful but currently there's only one way to unload quickly with one hand— by enabling the option to middle click tabs to unload. This issue is that I like to use middle click to close tabs. Right click and unload is not as fast. Ctrl + Middleclick requires two hands. 

So then you might say that middle clicking on the close button to unload is the best option? Almost. But the issue is that I can't do this on pinned tabs that have no titles because they have no close "x" button and even if they were added they would be to small of an area and sometimes you might accidentally close it instead of unloading it.

My solution would be for an option to unload if you middle click the tab's favicon. I mean you can already unload panels if you click their icon, so this seems intuitive. I also almost never want to close pinned tabs so middle clicking on them to unload only would be nice. I can also keep the functionality of closing normal tabs if I middle click to the right of the favicon. And most importantly it only requires one hand!

Issue #1304, opened 2023-10-02
Make pinned tab unclosable unless it's unpinned first


crowbardispensingcore:
## Description
This is the behavior in Safari. I think it makes more sense because you would want to avoid accidentally closing a pinned tab, which could happen when you use middle click to close tabs.

PR #1911, opened 2024-12-09
Add options for middle click on pinned tab

emvaized:
This pull request adds a separate configurable action for middle click on pinned tab. 

Related issues: 
https://github.com/mbnuqw/sidebery/issues/1304
https://github.com/mbnuqw/sidebery/issues/1320 

Considering it doesn't make much sense to close pinned tabs with middle click (as one of the reasons of the "pinned" state is to make tab not so easily closeable), this pull request makes "discard" a default action for the middle click on pinned tab. 

Available set of actions includes new option "unpin", which makes sense for the pinned tab, and excludes the option "duplicate as a child" (compared to the regular middle click configuration).


mbnuqw:
Yeah, sounds reasonable. Will merge it despite Firefox's default behavior (closing pinned tab on middle click), since Sidebery has distinguishable "unloaded" state of a tab. Thanks.

PR #1924, opened 2024-12-18
Added: option to show tab subtitle with URL

emvaized:
This pull request adds an option to show subtitle below each tab with its URL. It probably wasn't requested by anyone, but I suddenly came up with this idea, and after test-driving it for a few days I quite liked it. Seeing all urls at once allows to navigate tabs quicker, and makes it easier to distinguish between them.

By default this option is disabled, and should be manually discovered and enabled in Sidebery settings > Tabs. Subtitle shows url of each tab, but in the future it could possibly be used to show any other information. "https://" part of the URL is trimmed in the beginning, because I found it visually overwhelming (I'm not sure if "http://" should be trimmed as well).

This PR also adds some basic styles for the subtitle, such as smaller font and 70% opacity. It could probably also include some rules to increase default tab height when subtitles are enabled, but it currently doesn't do that, so tabs might become a bit too crowded with the default tab height. 

Here's how it looks with `--tabs-height=40px` set in the styles editor (recommended):

### DIFF
--- a/src/_locales/dict.setup-page.ts
+++ b/src/_locales/dict.setup-page.ts
@@ -2666,6 +2666,10 @@ Przykłady: "*", "ctrl+$", "ctrl+alt+g"`,
     zh_TW: '分頁切換間的最小延遲（毫秒）',
     ja: 'タブ切り替え間の最小遅延（ミリ秒）',
   },
+  'settings.tabs_show_subtitle': {
+    en: 'Show tab subtitle with URL',
+    ru: 'Показывать подзаголовок вкладки с URL',
+  },
 
   // - New tab position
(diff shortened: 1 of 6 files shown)


llc0930:
I feel that this feature is more closely related to the "Show URL in tooltip" option and should be placed before or after it.

emvaized:
> I feel that this feature is more closely related to the "Show URL in tooltip" option and should be placed before or after it.

Wouldn't this option then be misinterpreted as related to the tooltip?

llc0930:
I don't think so... Is the text of the two options hard to understand? Especially when they are already two different settings.

PR #1960, opened 2025-01-13
fix: search: should also search for custom titles for tabs

llc0930:
For example: I change the title of "google search" to "apple", then I expect apple should be able to retrieve the tab.


mbnuqw:
Yep, thanks.

PR #1961, opened 2025-01-14
fix: tab context menu: first sort the tree itself

llc0930:
Often when people click on the root of a tree, they want to sort the contents of the tree. I have a lot of tabs and after a few devastating sorts I've been using this setting.


mbnuqw:
Hm... indeed. I should also add an "Undo" notification for sorting

PR #1969, opened 2025-01-19
l10n: ja, zh: updated translation

llc0930:
I am not sure if and how Google translates the content in the Step 3 and Step 6 pages because I have not joined Google Workspace yet.
So 'settings.sync_gd_api_cli_sub' may be inaccurate.


llc0930:
Wait... I checked again, even if you choose "External", only people on the testing list can access it before the official release... I will revise the translation of that part.

llc0930:
Did you miss a step to add a new test user in https://console.cloud.google.com/auth/audience?

mbnuqw:
Thanks

> Did you miss a step to add a new test user

Yeah, forgot about it, will add it right now

PR #1985, opened 2025-01-26
fix: search: minimum search query length in CJK should be 1.

llc0930:
In short, the minimum search query length required for our writing system is 1.

The reason why CJK Unified Ideographs Extension B-I are not included is because those characters are too rare... I think they can pretty much only be entered using code points directly... and therefore it is unlikely to be used for naming and searching titles.

Let's look at a simple example in Traditional Chinese:
Monday = 週一 = 星期一
Tuesday = 週二 = 星期二
Wednesday = 週三 = 星期三
Thursday = 週四 = 星期四
Friday = 週五 = 星期五
Saturday = 週六 = 星期六
Sunday = 週日 = 星期日 = 星期天
When we want to search for each day of the week, one of the situations we might want to search by just "週".

### DIFF
--- a/src/services/search.actions.ts
+++ b/src/services/search.actions.ts
@@ -289,7 +289,12 @@ let query = ''
 let beforeSwitchingPanelId: ID | undefined
 export function search(value?: string, noSel?: boolean): void {
   if (value !== undefined) {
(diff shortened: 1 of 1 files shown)


mbnuqw:
Thanks

Issue #1991, opened 2025-02-04
Offer omnibox (address bar) handler for switching the current tab to a different container


ErichDonGubler:
### Description

Extensions like [Switch Container Plus](https://addons.mozilla.org/en-US/firefox/addon/switch-container-plus/) allow a user to very conveniently switch containers for the currently viewed tab. It's pretty great!

However, other extensions cannot be not aware of Sidebery's source of truth WRT tab placement, indentation, etc. From Firefox's perspective, the old and new tabs between containers are separate, with nothing to link an old tab to a new tab to an unrelated extension (which is a good default!). This means that when using them—as a continued example with Switch Container Plus, typing `co My Container`—Sidebery orders the new tab at the end of the sidebar, by default, losing the old tab's placement.

Sidebery should add a handler for the omnibox (keyed using a bikesheddable keyword like `sb`) to allow users to quickly change the container of their tab while keeping the placement of their tabs.

ErichDonGubler:
I've been test-driving #1992 myself to experiment with a solution for the last few months, and I've enjoyed it immensely. I'd encourage other folks to try it out.

ErichDonGubler:
#1992 has landed, so I see no reason to keep this open. Cheers!

PR #1992, opened 2025-02-04
Add `=` omnibox handler, use it for reopening tabs in a different container

ErichDonGubler:
Satisfies <https://github.com/mbnuqw/sidebery/issues/1991>.

### DIFF
--- a/src/services/omnibox.ts
+++ b/src/services/omnibox.ts
@@ -19,11 +19,14 @@ function setupListeners() {
   }
 
   browser.omnibox.onInputChanged.addListener(async (input, suggest) => {
-    const suggestions = matchContainers(input).map(ctx => ({
-      content: ctx.name,
-      description: ctx.name,
-      deletable: false,
-    }))
+    const suggestions =
+      input.length >= 3
+        ? matchContainers(input).map(ctx => ({
+          content: ctx.name,
+          description: ctx.name,
+          deletable: false,
+        }))
+        : []
     suggest(suggestions)
   })
 
--- a/src/services/omnibox.ts
+++ b/src/services/omnibox.ts
@@ -1,0 +1,71 @@
+import { Containers } from 'src/services/containers'
+import { translate } from 'src/dict'
+import * as IPC from 'src/services/ipc'
+import * as Logs from 'src/services/logs'
+import { Tabs } from 'src/services/tabs.bg'
+import { Windows } from 'src/services/windows'
+import { Container, InstanceType } from 'src/types'
+
+function setupListeners() {
+  browser.omnibox.setDefaultSuggestion({
+    description: translate('omnibox.container_switch.prompt'),
+  })
+
+  function matchContainers(input: string): Container[] {
+    // TODO: order by score of some sort?
+    return Object.values(Containers.reactive.byId).filter(container =>
+      container.name.toLowerCase().includes(input.toLowerCase())
+    )
+  }
+
+  browser.omnibox.onInputChanged.addListener(async (input, suggest) => {
+    const suggestions = matchContainers(input).map(ctx => ({
+      content: ctx.name,
+      description: ctx.name,
+      deletable: false,
+    }))
+    suggest(suggestions)
+  })
+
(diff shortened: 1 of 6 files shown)


ErichDonGubler:
I think this is basically ready for review, disclaiming that I'm not familiar with the vast majority of the Sidebery codebase. I haven't brought it out of draft yet, though, because the premise of this feature hasn't been validated by maintainership (see #1991).

mbnuqw:
Thanks, nice feature! Things that I'd change:
- Create service `omnibox.ts` and move that code to it. Export one function `setupListeners` and call it in `background.ts`: `omnibox.setupListeners()`.
- Use different keyword, something easily accessible from other, non-latin keyboard layouts. e.g. `=` (it seems that this symbol is available in all supported locales (de, fr, hu, pl, ru, zh, ja) without switching to en layout, although I could be wrong).
- Translate default suggestion

You can make it mergable and I'll complete it myself (prob also add switching between panels).

ErichDonGubler:
I've applied each of your suggestions. There are still several `TODO` comments that I'll call out as review discussions, but I think this can otherwise be reviewed normally! 😀

src/services/omnibox.ts
```diff
@@ -0,0 +1,73 @@
...
+import { Windows } from 'src/services/windows'
+import { Container, InstanceType } from 'src/types'
+
+function setupListeners() {
+  browser.omnibox.setDefaultSuggestion({
+    description: translate('omnibox.container_switch.prompt')
+  })
+
+  function matchContainers(input: string): Container[] {
+    // TODO: order by score of some sort?
```
ErichDonGubler, line 15: todo: Any suggestions on what to do here?

src/services/omnibox.ts
```diff
@@ -0,0 +1,73 @@
...
+    // TODO: At the very least, put an exact match first.
```
ErichDonGubler: todo: Any suggestions on what to do here?

src/services/omnibox.ts
```diff
@@ -0,0 +1,73 @@
...
+    // TODO: Validate we want Sidebery records, not Firefox records.
```
ErichDonGubler: todo: Any suggestions on what to do here?

mbnuqw:
Thanks

> order by score of some sort?

I though about sorting by the time of the last usage, but this will require adding logic for saving/loading info of the last N used commands. Too much for the initial implementation, so I think we can leave this as it is, a couple more of chars in the query will be enough for narrowing down the list of suggestions.

> At the very least, put an exact match first.

Not sure we need this right now, with only containers as the target. After all, containers usually have a different names? Later, when there will be more commands this can be added with the above mentioned sorting.

> Validate we want Sidebery records, not Firefox records

Not sure I understand this correctly. Sidebery mirrors the Firefox containers augmenting them with new features.

src/services/omnibox.ts
```diff
@@ -0,0 +1,73 @@
...
+    const activeTabs = sidebarTabs.filter(tab => tab.active)
+    try {
+      await IPC.sidebar(
+        Windows.lastFocusedWinId,
+        'reopenInContainer',
+        activeTabs.map(tab => tab.id),
+        firstMatchingContainer.id
+      )
+    } catch {
+      console.warn('failed to re-open tabs', activeTabs, 'in container', firstMatchingContainer)
```
ErichDonGubler: todo: Looks like I should use the `Logs` API instead.

ErichDonGubler:
@mbnuqw: I've eliminated the last two `TODO`s.

> … a couple more of chars in the query will be enough for narrowing down the list of suggestions.

This currently isn't implemented; did you want this in here before merging?

EDIT: Added a minimum limit of 3 for starting matching with c6a794bb64090be8633b154f98243564e4395cd0.

mbnuqw:
Thank you. Going to merge this.

> Added a minimum limit of 3 for starting matching

Sorry, (probably didn't convey my thoughts well) I'll revert this as it will make impossible for Japanese and Chinese users to search for one/two-length names of containers (in hieroglyphs). Such limitations are usually added with the large lists of source data that is not our case.

PR #2016, opened 2025-02-15
Add support for configuration through central policies

### LINKED ISSUE #1776: [Feature Request] Allow importing settings from a persistent directory in order to improve synchronization and reproducibility options
**Description**
I am trying to make my system (nixos) reproducible in order to make switching between devices and setting up new ones as seamless as possible. Unfortunately sideberry still requires some manual steps to keep in sync.

**Current Behavior:**
Afaik the only way to import all settings is using the "Import Settings" button in the extension settings UI. Firefox Sync does not seem to cover all settings and is therefore insufficient for syncing between devices and reproducibility for now. Regardless, an accountless way of syncing settings would be nice

**Requested Behavior:**
A persistently located directory to paste in a config and have it automatically be imported and applied. This would allow scripts to automatically configure the extension based on an earlier exported Sideberry config file from a git repo for example. 

Thank you for your work, Sideberry has been a great experience so far

spl3g:
This pr adds an option to configure Sideberry with a file in `~/.mozilla/managed-storage/ `or any other [manifest location](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_manifests#manifest_location).
It also resolves #1776

So for example if you want to set the sidebarCSS, you would need to create a file in a manifest location with something like that in the contents:
`~/.mozilla/managed-storage/{3c078156-979c-498b-8990-85f7987dd929}.json`:
```json
{
  "name": "{3c078156-979c-498b-8990-85f7987dd929}",
  "description": "_",
  "type": "storage",
  "data": {
    "sidebarCSS": "#root.root {--frame-bg: #343F44;}"
  }
}
```

Right now i added support for these keys:
- settings
- sidebarCSS
- groupCSS
- sidebar.nav
- sidebar.panels
- contextMenu

Any settings in the managed storage will overwrite the configuration in the local storage fully, except for the settings. It will overwrite only the fields that are in managed storage.


spl3g:
Hi @mbnuqw, is there anything preventing from merging the PR? Maybe I can improve or fix it.

src/services/settings.actions.ts
```diff
@@ -23,16 +23,23 @@ type Opts = typeof SETTINGS_OPTIONS
 export async function loadSettings(): Promise<void> {
   Logs.info('Settings.loadSettings()')
 
-  const stored = await browser.storage.local.get<Stored>('settings')
-
+  // The catch reason: https://bugzilla.mozilla.org/show_bug.cgi?id=1868153
+  const stored = (await browser.storage.managed.get<Stored>('settings').catch(() => {})) ?? {}
```
LucasWagler: Might want to rename for clarity. With current naming, it took me an extra moment to work through the `normalizeObject` logic below
```suggestion
  const storedManaged = (await browser.storage.managed.get<Stored>('settings').catch(() => {})) ?? {}
```

mbnuqw:
Thank you.

There are some UX issues (user should know what data is managed and cannot be permanently changed by Sidebery) but since this feature is for advanced users I'll merge it as it is.

PR #2017, opened 2025-02-16
Corrections in Polish

stankiewiczmichal:
I have made a few corrections to the Polish translation so that it conforms to the accepted nomenclature.


mbnuqw:
Thanks

Issue #924, opened 2023-02-03
Adding 'unload' to Double-Click tab actions


PuppyGirlBelly:
Hi, I recently switched to this extension after using Tree-Style tabs for almost a year. And I really enjoy what you've created, and it almost replicates my particular setup with minimal configuration. 

However, the one thing I am missing is the ability to unload tabs with a double click. Currently, the middle-click option allows for either closing or unloading. Which I appreciate, however I would prefer to not give up the option of easily closing a tab with a middle-click in order to easily unload a tab.

Here are the options on the current beta version I'm on (v5.0.0b31);

![image](https://user-images.githubusercontent.com/40257903/216671316-1264a89a-9257-4053-b291-f38f73a9b17a.png)

Thank you for your work, and I plan on using this extension for a very long time!

GodKratos:
I would really like this option too. Was surprised not to see it available under the double click selection.

the-nelsonator:
@PuppyGirlBelly @GodKratos any issues with unload on long-click (left or right) as an alternative? That's been given the thumbs-up by extension owner but unload on double-click won't be added until a tab-switch delay is added, see https://github.com/mbnuqw/sidebery/pull/2068#issuecomment-2764218479

GodKratos:
> [@PuppyGirlBelly](https://github.com/PuppyGirlBelly) [@GodKratos](https://github.com/GodKratos) any issues with unload on long-click (left or right) as an alternative? That's been given the thumbs-up by extension owner but unload on double-click won't be added until a tab-switch delay is added, see [#2068 (comment)](https://github.com/mbnuqw/sidebery/pull/2068#issuecomment-2764218479)

I'm not sure I fully understand what the problem is but I assume the issue is that if you double-click a tab that doesnt have focus, you will gain focus on the first click, and then lose focus when it unloads so the active tab will then change depending on the users settings when closing the active tab instead of staying on the tab they had active before any clicking happened.

TreeStyleTabs has an unloader extension to handle this behaviour which has a setting for the double-click delay.

![Image](https://github.com/user-attachments/assets/efb97031-24e9-47f4-b0ac-20eb82b95a90)

I assume the purpose of that delay setting is to add the "delay before activation" functionality the dev suggested in #2068 so I think that would be a viable way to implement it as it works well there. (The effect is that you unload unfocussed tabs without the current active tab changing)

GodKratos:
[...]

PR #2068, opened 2025-03-28
Add unload to long-click tab actions

the-nelsonator:
Relates to https://github.com/mbnuqw/sidebery/issues/924

image

Reuses existing `settings.tab_action_discard` in `dict.setup-page.ts`


mbnuqw:
Thanks. While long-left/right-click options are ok and mergable, unloading with double click will lead to double-switching active tab and the final active tab might differ from the initial (depending on user's settings), which I consider as an unexpected behavior (plus that screen flash from fast tabs switching).

Ideally, when this option is used, sidebery should add a delay before tab activation. And this delay should be configurable, since we don't know double-click timings. We also can shorten this delay and activate tab right after mouseleave event.

So to merge this we either need to remove double-click option (maybe it'll be enough to use long-click actions instead of double-click for people from #924...) or add a deferred tab activation mechanism with delay.

the-nelsonator:
Long-click definitely seems to be more usable for unload. Went with simpler option and just removed double-click

the-nelsonator:
@mbnuqw actually, am I misunderstanding your concern or doesn't left long-click face the same issue? At least the "double-switching active tab" part. Tab switches on mouse down, then when long press triggers it unloads so you switch to another tab. Do you see this behavior as bad enough where it should just be limited to right long-click? Or is it fine because long-click is slower than double-clicking?

mbnuqw:
> doesn't left long-click face the same issue?

Yes, you're right, I forgot about this case. There is a setting named "Activate tab on mouse button release" (in Mouse/Tab actions section), that needs to be enabled to avoid this behavior. Like in case with the 'edit_title' action here: https://github.com/mbnuqw/sidebery/blob/13ea8b329746d9cd4cef75ea67a341e9bd5b99d8/src/page.setup/components/settings.mouse.vue#L397-L402. I'll fix this.

Issue #2067, opened 2025-03-25
[Feature Request] Add option to "Show Container in Tooltip"


GodKratos:
### Description

There is an option to "Show URL in Tooltip" can we also get an option to "Show container in tooltip" please?
Sometimes I have tabs open in different containers with the same (or similar) colour and I'd like to tell which container the tab is open in when I hover over the tab.

PR #2075, opened 2025-04-01
feat: option to include container name in tab tooltip (#2067)

GodKratos:
I've added the feature to include the Container name in the tab tooltip for #2067 
If a tab has no container or is using the default it does nothing, but adds the container name when a container exists.

image

image


mbnuqw:
Thank you

PR #2079, opened 2025-04-02
Normalize bundle path slash directions so building on Windows works

### LINKED ISSUE #2078: [Dev build issue] Error when opening tab group page, script doesn't inject
### Steps to reproduce
- Reset sidebery repo (to rule out any untracked files being related)
  - `git clean -f -d -x`
- Run sidebery in dev build mode following readme
  - `nvm use 22.12.0` (latest LTS)
  - `npm install`
  - `npm run dev`
  - (In another terminal tab) `npm run dev.run -- {firefox install location}`
- Open a tab group page

### Actual behavior

Displays a default page

Image

### Expected behavior

Should display the tab group page in selected mode

### System

Windows 11

### Firefox version

138.0b1

### Sidebery version

v5.3.3

### Logs

```code
SyntaxError: import declarations may only appear at top level of a module group.js:2:1
[bg] Tabs.injectGroupPageScript: Cannot inject script, tabId: 3 Error: import declarations may only appear at top level of a module logs.ts:27:10
```

Compiled group.js: 

Image

Not sure yet whether it's "use strict" or something with the import format or something else causing the issue. I assume this is environment dependent or @mbnuqw would have ran into this when developing the tab group page!

the-nelsonator:
Closes https://github.com/mbnuqw/sidebery/issues/2078

`scripts.js#BUNDLES` uses Unix-style forward slashes for filepaths:

```
17:46:45 {
  'src/injections/group.ts': true,
  'src/injections/url.ts': true,
  'src/injections/tab-preview.ts': { format: 'iife' },
  'src/popup.tab-preview/tab-preview.ts': true
}
```

These are checked in `#compileTSFile` to determine whether to bundle or not.

But on Windows, the filepaths from `#getSrcFiles` have backslashes:

```
17:46:45 src\injections\group.ts
```

So the files aren't bundled and they can't be injected, hits `Error: import declarations may only appear at top level of a module` when trying to inject a bundled script - for example when opening a group page.

This uses the normalized Unix-style path when checking in  `#compileTSFile` which seems to resolve the issue.

### DIFF
--- a/build/scripts.js
+++ b/build/scripts.js
@@ -39,7 +39,7 @@ function fixModuleImports(data) {
 }
 
 /**
- * Replace Windows-style path separators with Unix-style paths
+ * Replace Windows-style path separators (\) with Unix-style (/)
  */
 function normalizePath(p) {;
   return p.replace(/\\/g, '/')
--- a/build/scripts.js
+++ b/build/scripts.js
@@ -41,7 +41,7 @@ function fixModuleImports(data) {
 /**
  * Replace Windows-style path separators (\) with Unix-style (/)
  */
-function normalizePath(p) {;
+function normalizePath(p) {
   return p.replace(/\\/g, '/')
 }
 
--- a/build/scripts.js
+++ b/build/scripts.js
@@ -39,6 +39,13 @@ function fixModuleImports(data) {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


mbnuqw:
Thanks

PR #2081, opened 2025-04-02
Hide preview box of unloaded tabs

### LINKED ISSUE #2080: [Enhancement] Remove blank tab preview from preview popup for unloaded tabs
### Description

Currently displays a blank section:

Image

This just takes up space and should be removed for unloaded tabs, leaving only the title and url.

the-nelsonator:
Closes https://github.com/mbnuqw/sidebery/issues/2080

When hovering over an unloaded tab with tab preview turned on (either in-page and popup):

Before: preview box is large blank area

429318394-ee680500-b453-4c2d-8dac-9af1c8bb3203

After: preview box not shown

image

Note: removed the bottom border radii to make it look more consistent when moving mouse over unloaded and loaded tabs. Let me know if you think it looks better without this.

image

Note: I think there might be a minimum window height, I couldn't figure out how to make the popup smaller if it's possible.


the-nelsonator:
@mbnuqw semi-related question: are there any benefits of the popup preview over the in-page preview? In-page seems cleaner with no downside or difference in functionality that I've noticed

mbnuqw:
Thanks.

I'll leave border radius of the bottom corners to align with the native implementation (plus, it's subjectivelly looks nicer :3).

Regarding the window-popup mode: It's actually a subject to remove, since it has issues in most environments (because it's just a window, that's stealing the focus from the main window). It also has default window interface (removable with userChrome). At the same time the 'in-page' mode has main three disadvantages:
- doesn't work on some pages (internal, mozilla's, other addon's) and need some fallback
- if the current page is heavy, it might seriously lag
- might not work right when fingerprinting protection is enabled (e.g. #2139)

PR #2083, opened 2025-04-02
Add setting to include active tab on initial ctrl+click

### LINKED ISSUE #2082: [Enhancement] setting to make ctrl+click also add active tab to the selection
### Description

With the following situation:

- Tab A (active)
- Tab B

When you ctrl+click on Tab B, currently just Tab B is selected. In vanilla FF tabs and other software with tabs, both Tab A and Tab B are selected, so a setting should be added to choose between current and "normal" behavior.

There's already a setting for shift+click: 

Image

the-nelsonator:
Depends on https://github.com/mbnuqw/sidebery/pull/2079 (I can split if needed, just can't build without it)

Closes https://github.com/mbnuqw/sidebery/issues/2082

image

With the following situation:

- Tab A (active)
- Tab B

When you ctrl+click on Tab B:

- Setting true: both Tab A and Tab B are selected. 
- Setting false: just Tab B is selected. 

Checks that there's not an active selection so it will only include active on initial ctrl+click.

Defaulted to true, since it seems to be the standard behavior in other software including vanilla FF tabs. But Sidebery users may be used to current behavior, so we can change to default false if that's preferred. 

First time adding a setting, I just plugged the English into Google Translate for the i10n so please let me know if there's a different process to follow!


build/scripts.js
the-nelsonator, line 1: I'm not sure why it's still showing changes to this file after https://github.com/mbnuqw/sidebery/commit/b7946ae83ea1477e08ef21328e6d019f80a04669, they're no-changes now. Maybe takes a while to update? I've synced the fork

the-nelsonator: Whoops, just had to pull into branch

mbnuqw:
Thank you!

PR #2114, opened 2025-05-01
Support using shift after clicking the first tab to open in a new window

zachswasey:
Currently a Shift+Click in the Snapshot view only works if Shift was held when clicking the first checkbox, contrary to many other implementations of the same mechanism. This makes it so Shift can be held after a single checkbox is selected, to select a range.

It took me an embarrassingly long time to realize that the Snapshot view even supported Shift range selection, because I'm not used to having to hold Shift on the first click since most other implementations don't require that.


mbnuqw:
Thank you

PR #2116, opened 2025-05-04
Support opening Snapshots in private windows

zachswasey:
I've been getting annoyed that snapshots can't easily be opened in private windows, even if the tabs were initially in a private window. At the moment a user has to open a private window manually and then select to open tabs in the current panel. This change gives the option to open windows in a private window, or open tabs in either the current panel, a new window, or a private window. Preferably the snapshot itself would also maintain the incognito state, which I'd also like to do, but that seemed like a much bigger change that I haven't explored fully yet.


mbnuqw:
Thanks

PR #2150, opened 2025-06-04
feat: add option to disable auto-scroll to new tab

### LINKED ISSUE #1270: Option regarding scrolling to position of new tab
In the older version of Sidebery, opening a new tab wouldn't cause the sidebar to automatically scroll to the position of the new tab. In 5.0, opening a new tab now does that. I can't find any option to disable this in the settings.

ingjieye:
(resolves #1270)

When many tabs are already open and more inactive tabs continue to be opened over time, this option helps keep the scroll position fixed, allowing users to process tabs one by one in order, instead of constantly being auto-scrolled to the newest inactive tab.


mbnuqw:
Thank you.

PR #2156, opened 2025-06-08
l10n: zh: jp: updated translation

llc0930:
Regarding commit 07718a9 and commit cb9d850, this feature currently only deduplicates the selected tabs.
I only made corrections to the Chinese and Japanese translations of `menu.dedupe`.


mbnuqw:
Thanks. I removed the "selected" part because it seems to me obvious that when the user selects the tabs and opens the context menu all context menu options are targeted at that selection and if they're not additional clarification should be added (e.g. options "Close" and "Close tabs above", "Close tabs below", etc...). But I may be wrong about how obvious this is, so if there are open issues related to this I'll consider adding clarifications to other options. For now I'll leave other langs for this option as they are.

Oh, just noticed that this dedupe option is not inactivated when only one tab is selected, which might be the reason for confusion, will fix this.

llc0930:
> Oh, just noticed that this dedupe option is not inactivated when only one tab is selected, which might be the reason for confusion, will fix this.

Yes, because only one tab is selected and it is not inactivated at the time.

llc0930:
But perhaps people will just wonder: "Why can't this function be used after being added to the menu?" Perhaps it still needs to be modified?

PR #2167, opened 2025-06-14
l10n: add possibility to translate all headers in Styles editor

ariasuni:
Note: removed “Frame colors” and “Toolbar colors”, they seem unused because they are captured in the “General” category.


src/_locales/dict.setup-page.ts
```diff
@@ -6558,6 +6558,34 @@ Przykłady: "*", "ctrl+$", "ctrl+alt+g"`,
...
+    en: 'General',
+    fr: 'Général',
+  },
+  'styles.vars_group.history': {
+    en: 'History',
+    fr: 'Historique',
+  },
+  'styles.vars_group.popup': {
+    en: 'Popup',
+    fr: 'Popup',
```
llc0930, line 6571: If it's the same, there's no need to force yourself to write it twice.

src/_locales/dict.setup-page.ts
```diff
@@ -6558,6 +6558,34 @@ Przykłady: "*", "ctrl+$", "ctrl+alt+g"`,
...
+    en: 'Search bar',
+    fr: 'Barre de recherche',
+  },
+  'styles.vars_group.new-tab_bar': {
+    en: 'New-tab bar',
+    fr: 'Barre de nouvel onglet',
+  },
+  'styles.vars_group.notification': {
+    en: 'Notification',
+    fr: 'Notification',
```
llc0930, line 6587: the same

mbnuqw:
Thanks

PR #2171, opened 2025-06-16
l10n: updated translation

llc0930:
Perhaps for the last time, I don’t plan to open an account on any localized platform.


mbnuqw:
Thanks.

> open an account on any localized platform

That's definitely won't be required.

Issue #2161, opened 2025-06-10
“Sync sub-panel” in setup page isn’t translatable


ariasuni:
### Steps to reproduce

“Sync sub-panel” in setup page isn’t translatable.

### Actual behavior

Can’t translate “Sync sub-panel”.

### Expected behavior

Can translate “Sync sub-panel” with Sidebery’s translate workflow.

### System

Arch Linux

### Firefox version

139.0.1

### Sidebery version

v5.3.3.17 nightly

### Logs

```code

```

PR #2172, opened 2025-06-16
l10n: Add ability to translate some strings with missing translation

hannaeko:
Make some strings translatable:
* Addresses #2161 by adding the ability to translate entries titles and success notifications in the sync sub panel in the setup page.
* Add the ability to translate the error message when the permission for private navigation is not granted when reopening a tab in a private window.
* Make “Not set” option in auto move rule translatable.


ariasuni (approved):
I checked the French translations and they’re 💯. The code changes seem fine but I didn’t look too hard.

mbnuqw:
Thanks

PR #2180, opened 2025-06-29
feat: Option to config automatic tab-panel scrolling

### LINKED ISSUE #260: Do not scroll to previously active tab when closing active tab
My apologies about the cryptic title. My request refers to the following situation:

- I have more tabs open in a panel than fit in the window (so there is a scroll bar)
- I have "after closing current active tab activate" set to "previously active tab"

When I close the currently active tab and the previously active tab is on the opposite side of the panel, the previously active tab is not only activated, but the panel also scrolls to that tab.

To me, the latter behaviour is quite annoying because it frequently causes me to misclick when I want to activate/close another tab right after closing the active tab.

It would already be a big improvement if the scroll was animated (it does not currently appear to be), but actually I would welcome an option that turns off the scrolling behaviour completely, so the tab is only activated but not scrolled to.

deviant:
This is more or less identical to https://github.com/mbnuqw/sidebery/commit/d94478bf34f34b06c4fba7419ffe946c664b62b3, except instead of controlling panel switching, it affects panel scrolling behaviour.

Fixes #260.


deviant:
@mbnuqw Would it be possible to take a look at this? Thanks!

mbnuqw:
Thanks

PR #2213, opened 2025-08-05
fix: Search popup sometimes not focusing

### LINKED ISSUE #1810: Search doesn't work
### Steps to reproduce

- Install Sidebery
- Set hotkey to search tabs
- Use hotkey

### Actual behavior

Either doesn't open search, even when the hotkey isn't reserved (mostly on builtin pages like new tab page) or if it does open, it's not focused

### Expected behavior

Press hotkey and it focuses on search

### System

Windows 11 Pro

### Firefox version

130.0.1

### Sidebery version

5.2.0

### Logs

_No response_

Ascyt:
Fixed #1810. It seems to focus mostly consistently for me now. I've still had a handful of cases where it still didn't focus, however it seems to work much better than the current version.

Changes in this PR:
- Try to focus the element 10 times in the first second
- Sidebery's default search bar now only shows when in focus of sidebery. Previously, it would show it even when sidebery was not in focus. I believe this primarily led to the issue, as I suppose the search window of sidebery also tried to get focus, which conflicted.
- Removed the 300ms delay for search. Not sure why this is a thing in the first place, it works perfectly fine without the delay and is much more snappier and usable, especially when using this feature a lot and frequently switching between tabs.

### DIFF
--- a/src/popup.search/search.ts
+++ b/src/popup.search/search.ts
@@ -10,6 +10,12 @@ const el = document.getElementById('textInput') as HTMLInputElement
 
 el?.focus()
 
+for (let i = 0; i <= 500; i += 100) {
+  setTimeout(() => {
+    el?.focus()
+  }, i)
+}
+
 el?.addEventListener('blur', () => {
   if (Windows.id !== undefined) IPC.sendToSidebar(Windows.id, 'onOutsideSearchExit')
 })
@@ -92,6 +98,7 @@ void (async () => {
   IPC.setInstanceType(InstanceType.search)
   IPC.setupGlobalMessageListener()
   IPC.registerActions({ closePopup })
+  
   const [win] = await Promise.all([
     browser.windows.getCurrent({ populate: false }),
     Settings.loadSettings(),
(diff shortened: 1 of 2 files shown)


Ascyt:
Quick question @mbnuqw , is there really any need for the 300ms delay when searching? From what I'm seeing it cannot be disabled in the settings, and it would be nice to use a shortcut to quickly switch around named tabs without having to wait a third of a second. I'm a fast typist and the delay is very hindering. In this PR I removed the delay completely and didn't encounter any issues without it, and it all worked much better and felt really snappy. Would be cool for this to be removed or at least some option to disable the delay in the settings.

mbnuqw:
> is there really any need for the 300ms delay when searching?

Yes, on potato-grade hardware (like mine) it's near-impossible to search without this debounce in the large lists of elements (tabs and especially bookmarks). But I'm planning to remove it eventually. First, for the small amount of tabs (<100), then, when I'll optimize rendering, for everything.

Right now, you can remove this delay by changing `searchInputTimeout` hidden pref: Open Sidebery settings, change some setting (and revert it, just to make sure that fresh settings are saved b/c it's a recent pref). Then in 'General' section press on the 'Sidebery's data: ...' button, press on the 'settings' in the displayed list and then on 'EDIT' button in the top-right area. Find `searchInputTimeout` and change its value (300) to e.g. 0. Press 'SAVE' in the top-right area. Done.

PR #2215, opened 2025-08-08
Update Firefox-Styles-Snippets-(via-userChrome.css).md

salattu:
Added step to additionally hide the 'revamped sidebar' with Firefox tools. The instructions before this addition left a sidebar further to the side of Sidebery.


mbnuqw:
Thanks

Issue #10, opened 2019-03-03
Other Platforms


lelpofi89:
Hello. Is it somehow possible to get Sidebery on Chrome? Thanks.

mbnuqw:
No, unfortunately Chrome doesn't support sidebar and contextual identities WebExtension APIs, so there are no way (at least simple) to build this addon for Chrome

emvaized:
@mbnuqw 
And what about Vivaldi browser? It's built on Chromium, and supports sidebars, even custom ones.

code3z:
> @mbnuqw
> And what about Vivaldi browser? It's built on Chromium, and supports sidebars

No, it doesn’t support sidebars for extensions. Opera might tho.

emvaized:
@code3z 
You are mistaken — any extension's page could be set as a sidebar in Vivaldi in just a few steps. 
And there's even already existing tree-style-tabs extension as an example (abandoned though): https://forum.vivaldi.net/topic/15332/tree-tabs

mderazon:
Do you know how these guys are doing it ?
https://getforest.io/

code3z:
@mderazon Probably with a script injected into the webpage.

@emvaized Yeah, that’s not an API though. If you want to add support for Vivaldi and can do it in a simple way and maintain it yourself, the author would probably accept it.

emvaized:
@code3z 
Actually I remember asking Sidebery's developer about this some time ago, and he said that in V5 he may try to reduce dependence on Firefox-specific APIs, in order to make it more compatible with Vivaldi. However I'm not sure whether this was done or not – currently any attempts to install latest beta release fail on stage of reading the manifest file.

solarkraft:
Chromium actually has a side panel that can be tweaked to work a lot like Firefox's now (flags: unified-side-panel, side-panel-web-view, settings: appearance > side panel)! All that's missing now is a good tree style tabs extension!

Considering the features they're adding regarding the side panel I'd say there's a good chance they'll eventually natively support showing extension panels, so now is the best time to reconsider a port!

> Do you know how these guys are doing it ?
> https://getforest.io/

One way they use (I think there are several, injecting into pages may be one of them) is to open an extra window or a pop-up when the extension's icon is clicked. The view that is loaded in these cases can be loaded into the Chromium side panel web view without any issues.

solarkraft:
[...]

PR #2234, opened 2025-09-04
Manifest v3

### LINKED ISSUE #916: Manifest V3 migration

teidesu:
the previous pr #1608 for some reason stalled and was never merged. 
i too am interested in seeing sidebery working under chrome :3

i picked up after them and rebased, though it seems as if most of the issues you raised in that pr have already been addressed
i only really updated the bg script to set up listeners in top-level to (hopefully) work correctly with mv3 event pages

a few open questions i still have:

- how to test that non-persistent background script works fine? i.e. what features rely on the background script being persistent?
- "Record state changes in local storage" – i *think* it's already storing everything it needs in local storage? not sure
- "Change timers to alarms" – seems like there aren't any long-running timers in the background script?
- "Place menu creation [...] in a `runtime.onInstalled` listener" – haven't really changed anything there, but menus *seem* to be working fine

---

if we're going the "chrome support" route, we'll eventually need to get rid of `webRequest`/`webRequestBlocking` api usages too.
i don't entirely understand where they are currently used, seems like something related to containers..? if so, can probably be skipped for now as chrome doesn't have anything like that

---

will close #916

### DIFF
--- a/src/bg/background.ts
+++ b/src/bg/background.ts
@@ -122,7 +122,7 @@ void (async function main() {
 function initToolbarButton(): void {
   Menu.createSettingsMenu()
 
-  browser.browserAction.onClicked.addListener((_, info): void => {
+  browser.action.onClicked.addListener((_, info): void => {
     if (info && info.button === 1) browser.runtime.openOptionsPage()
     else browser.sidebarAction.toggle()
   })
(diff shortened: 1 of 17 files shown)


the-nelsonator:
Some ports to Chrome have been linked in this thread: 

https://github.com/mbnuqw/sidebery/issues/10#issuecomment-4150129333

https://github.com/mbnuqw/sidebery/issues/10#issuecomment-5016129198

PR #2245, opened 2025-09-22
l10n: updated translation

llc0930:
**dict.setup-page.ts**

`panel.tab_move_rules_manage_badge`
Is it still in use?

`panel.move_rules_active`
`panel.move_tab_ctx`
`panel.move_tab_ctx_nochild`
`panel.url_label`
`panel.url_rules`
These five appear to be no longer in use.

`settings.scroll_panel_after_switching_tab`
Waiting for confirmation of actual functionality to correct the translation.


llc0930:
About `settings.scroll_panel_after_switching_tab`:
https://github.com/mbnuqw/sidebery/issues/260#issuecomment-3285992826

mbnuqw:
> `panel.tab_move_rules_manage_badge`
> Is it still in use?

Yep, it's a tooltip for the button.

> ...
> These five appear to be no longer in use.

Yes, it seems so, can be removed

> About `settings.scroll_panel_after_switching_tab`

Ouch... yeah, it should be something like 'Automatically scroll panel to activated tab'

llc0930:
> > `panel.tab_move_rules_manage_badge`
> > Is it still in use?
> 
> Yep, it's a tooltip for the button.

Well, while I know it is indeed used, I can't really find where it is shown.

mbnuqw:
> Well, while I know it is indeed used, I can't really find where it is shown.

Settings > Navigation bar section > Enabled elements > Any tab panel with more than 0 auto-move rules > A button/badge with arrow icon and count of rules

Thanks

llc0930:
Ah, it turns out to be the tooltip for the new button on the settings page. No wonder I couldn’t find it on the rule management page.

Issue #2170, opened 2025-06-16
Better and easier localisation


ariasuni:
### Description

Currently, translating Sidebery is a pain: no context for strings, you have to know Git and handle conflicts when stuff is pushed after you’ve started translating locally, some code use concatenation that only works with English word order, etc. And since Sidebery has A LOT of strings, it’s really necessary IMO to have a better translation workflow.

Ideally:

- [ ] Replace concatenation logic by something that can be properly translated (I’m preparing a PR for that, [here’s the branch](https://github.com/ariasuni/sidebery/commits/dont-concatenate-translation-strings/))
- [ ] Use standardized formats instead of the current custom one, with e.g. https://github.com/projectfluent/fluent.js/ or https://formatjs.github.io/docs/intl-messageformat/, that can handle plural rules by language automatically and use variables.
- [ ] Use a Weblate or something to manage localisation (other advantages: easier to avoid forgetting strings or forgetting punctuation at the end of string, etc.)

mbnuqw:
> no context for strings

Comments can be used for this. 

> Replace concatenation logic by something that can be properly translated (I’m preparing a PR...

Agreed, this should be done, but I'd use different PlurFn type (the name should also be updated since these functions have more general usage now):
```ts
/// globals.d.ts
type PlurFn = (...args: (number | string)[]) => string

/// dict.ts
function translate(id?: string, ...args: (number | string)[]): string {
  //...
  else return record(...args)
}

/// dict.sidebar.ts
'newTabBar.new_tab_in_container_with_url': {
  en: (url, name) => `Open ${url} in a new tab in "${name}" container`,
},
```

> Use standardized formats instead of the current custom one...

Meh. Lots of work (probably), while the current approach covers all Sidebery needs, it's simple, performant and flexible (with function props). Plus, I'd like to keep as few as possible runtime dependencies (vue).

> Use a Weblate or something to manage localisation...

Not against it, I'll check it a bit later.


I also plan to separate translations into dirs by language to decrease count of merge conflicts.

PR #2250, opened 2025-09-25
l10n: don’t concatenate translation strings, use more source strings

ariasuni:
Fixes part of #2170.

I totally eliminated the complicated logic around the tooltips for new tab buttons and instead created a bunch of strings to cover each case and greatly ease translatability (the strings are slightly different because I harmonized them a little bit).

Otherwise, it’s all pretty obvious stuff, and I did my best to not alter current behavior.


mbnuqw:
Thank you.

PR #2256, opened 2025-10-02
l10n: updated translation

llc0930:
`newTabBar.open_child_tab_in_default_container`
`newTabBar.open_child_tab_in_default_container_with_url`
The child tab will open in the same container as the parent tab, which is different from what the variable names describes. Does the current behavior meet the correct goal?

### DIFF
--- a/src/_locales/dict.setup-page.ts
+++ b/src/_locales/dict.setup-page.ts
@@ -458,7 +458,7 @@ Liste de “sous-chaines” ou “/RegExp/” avec une entrée par ligne:
     hu: 'További „Új lap” parancsikonok',
     pl: 'Dodatkowe skróty "Nowa karta"',
     ru: 'Дополнительные кнопки для создания новой вкладки',
-    zh_CN: '额外的"新建标签页"快捷方式',
+    zh_CN: '额外的"新标签页"快捷方式',
     zh_TW: '額外的「新分頁」捷徑',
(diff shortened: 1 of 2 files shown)


mbnuqw:
It's a bug (alright... it's just poorly designed feature). Presently 'Default container' option (in new-tab-button shortcuts) working as 'Not set' for the new tab. This results in inheriting container from the parent tab (default Firefox behavior when cookieStoreId is not set). I should update new-tab-shortcuts editor and new-tab-button components: add a 'Not set' container option (for the current behavior) and update 'Default container' option so that the container for child tabs is always default.

Current translations(en) are technically alright, (they doesn't say that child tab will be open in the default container), though mentioned property names will be updated: `newTabBar.open_child_tab_in_default_container` > `newTabBar.open_child_tab`, ` newTabBar.open_child_tab_in_default_container_with_url` > ` newTabBar.open_child_tab_with_url`.

llc0930:
Well, I was actually just asking if the variable name should be changed, since the actual behavior is different. The translation and behavior are the same, although some translations don't explicitly mention the 'Default container', but that's okay, I understand.

Are you going to update the variable names yourself when you fix this feature, or should I do it for you? It seems like this should be split into its own commit?

mbnuqw:
> Are you going to update the variable names yourself when you fix this feature, or should I do it for you?

I'll do this myself

PR #2260, opened 2025-10-07
l10n: updated translation

llc0930:
Firefox's own translation in `aboutSessionRestore.ftl` was not very consistent, so in the end I chose a relatively compromise translation.


mbnuqw:
Thanks

PR #2274, opened 2025-10-20
l10n: updated translation

llc0930:
Maybe I should accumulate more and then do a PR before a new Nightly release.


mbnuqw:
> Maybe I should accumulate more and then do a PR before a new Nightly release.

Thanks. Do as you feel comfortable. This week I plan to go through latest github issues fixing some interesting bugs (prob with a couple of additional nightly releases) and on the weekend make v5.4.0 for AMO.

PR #2282, opened 2025-11-01
l10n: zh: updated translation

llc0930:
It seems you were referring to this week last time. So maybe I shouldn't have let this drag on until the end of the weekend.


mbnuqw:
Thanks. No, it's just that bugs keep popping up here and there :sweat_smile:, which happens before each AMO release b/c I test features I rarely use (it seems soon I'll start to add unit and e2e tests...). So if in the future I mention some deadline just increase that duration x2 or x3.

PR #2291, opened 2025-11-11
l10n: fr: fix the translation for 'Unload'

cluxter:
'Unload' was translated as 'Actualiser', but 'Actualiser' means 'Refresh' in English. 'Unload' can be translated as 'Décharger' in French.


mbnuqw:
Thanks

Issue #2175, opened 2025-06-20
Option that drag-and-drop tab to a different panel will put the tab at the top of tab list


VictorVow:
### Description

Currently it defaults to the bottom of the tab list.

PR #2317, opened 2025-11-30
Add moveTabToPanel setting for configurable tab drop position

VictorVow:
Implements #2175: Users can now configure where tabs land when dragged to a panel.

## Changes

**New setting**: `moveTabToPanel` in Drag and Drop settings
- Options: `'start'` | `'end'` (default: `'end'`)
- Controls tab placement when dropped to panel area (not between specific tabs)

**Modified `getDstIndexInside()` logic**:
```typescript
// Before: always used nextTabIndex (end of panel)
return dstPanel.nextTabIndex ?? Tabs.list.length

// After: respects user preference
return Settings.state.moveTabToPanel === 'start'
  ? dstPanel.startTabIndex
  : (dstPanel.nextTabIndex ?? Tabs.list.length)
```

**Files changed**:
- `src/types/settings.ts` - Type definition
- `src/defaults/settings.ts` - Default value and options
- `src/page.setup/components/settings.dnd.vue` - UI control
- `src/_locales/dict.setup-page.ts` - i18n (9 languages)
- `src/services/drag-and-drop.actions.ts` - Drop logic

Precise positioning (dropping between specific tabs or onto parents) is unaffected.


mbnuqw:
Thanks

PR #2332, opened 2026-01-01
feat: setting field improvements

### LINKED ISSUE #2275: [Enhancement] Make settings labels selectable, clicking not toggle/select setting
### Description

For ease of copying settings labels for reference to others, their text should be selectable and the entire section shouldn't act as an input for the setting - only the dropdown/toggle should be clickable.

For example, right now I can't highlight and copy "Double click on tab" - trying to select that text just activates the dropdown:

Image

Best workaround is inspect element in the meantime

FurkanKambay:
- Resolves #2275
- Fixes dropdown overlay click-through bug
- Prevents context menus opening when right-clicking on a field, except when text is selected
- Fields with two inputs now focuses the `TextInput` on left-click, and switches/toggles the `SelectInput`/`ToggleInput` on right-click.

https://github.com/user-attachments/assets/9bfb67b1-2d75-4e14-be4d-6a17e39783bd

### DIFF
--- a/src/styles/inputs.styl
+++ b/src/styles/inputs.styl
@@ -28,7 +28,6 @@
   padding: 0 16px
   flex-shrink: 0
   cursor: default
-  -moz-user-select: none
   outline: none
   border-radius: 4px
   &:not(:first-of-type)
@@ -197,12 +196,16 @@
   padding: 0 0 12px 0px
   white-space: pre-wrap
   color: var(--toolbar-fg)
-  user-select: none
   opacity: .7
 
 // ---
 // -- Input
 // -
+.ToggleInput
+.SelectInput
+.TextInput
+  -moz-user-select: none
+
 .ToggleInput
 .SelectInput
   position: relative
@@ -415,6 +418,7 @@
   position: relative
   padding: 2px 3px
   overflow-y: auto
+  user-select: none
   scrollbar-color: var(--frame-scrollbar-color) #0000
 #root[data-native-scrollbars-thin="true"] .select-input-drop-down-content
   scrollbar-width: thin
@@ -650,7 +654,6 @@
     align-items: center
     color: var(--toolbar-fg)
     transition: color var(--d-fast)
-    -moz-user-select: none
   .input-group
     position: relative
     display: flex
@@ -706,7 +709,6 @@
 .StyleField
   position: relative
   padding: 8px 16px 2px
-  -moz-user-select: none
   &:not(:first-of-type)
     margin: 3px 0 0 0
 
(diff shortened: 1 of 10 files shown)


mbnuqw:
Thanks, but I'd like to be able to switch settings by clicking on the whole field area. This can be achieved by checking the current selection's type in mousedown/up event handlers of the top-most `x-field.vue` element.

Something like this:
```ts
let rangeIsSelected = false
function onMouseDown(e: DOMEvent<MouseEvent>) {
  rangeIsSelected = getSelection()?.type === 'Range'
  if (e.detail > 1) e.preventDefault()
}

function onMouseUp(e: DOMEvent<MouseEvent>) {
  if (rangeIsSelected || getSelection()?.type === 'Range') return

  // And then field-specific stuff:
  // For select-field
  if (props.inactive || !props.opts || Array.isArray(props.value)) return
  if (e.button === 0) switchOption(1)
  if (e.button === 2) switchOption(-1)
  // For toggle-field and count-field
  toggle()
  // text-field
  focus()
}
```

No need to change select-input component in this case.

src/styles/inputs.styl
mbnuqw: This line can be removed, so user can select notes too.

FurkanKambay:
@mbnuqw perfect, thank you for the direction!

just wanna take a look into a couple more small improvements before its ready to merge :)

FurkanKambay:
I see that `NumField`s don't have mouse handlers because they can contain two input parts like this

image

Should it handle mouse events when the `SelectField` is hidden, or maybe LMB focuses the `TextInput` and right click modifies the `SelectInput`?

Similarly, the `Limit the count of simultaneously reloading tabs` setting (`CountField`) has both a `TextInput` and a `ToggleInput`, but clicking it just toggles it. I think left click to focus and right click to toggle would work really nice for these cases.

**UPDATE:** I implemented this (and for `StyleField`s) in the latest commit

FurkanKambay:
Also fixed this click-through propagation bug:

https://github.com/user-attachments/assets/9bfb67b1-2d75-4e14-be4d-6a17e39783bd

FurkanKambay:
sorry for the messy force push stuff, I wanted the commits to be tidier.
Almost ready for review now, I can revert the latest commit if you don't like that behavior (https://github.com/mbnuqw/sidebery/pull/2332#issuecomment-3761621808)

~~caught a bug, working on it~~

The fields in `PanelConfig` have a weird text selection bug, couldn't figure it out yet.

mbnuqw:
Awesome, thank you!

> Should it handle mouse events when the SelectField is hidden, or maybe LMB focuses the TextInput and right click modifies the SelectInput?

Yes, let's keep you're implementation.

> The fields in PanelConfig have a weird

That's `selectionchange` handler in popup.panel-config/panel-config.vue. I'll fix this after merge.

PR #2337, opened 2026-01-15
fix: middle click to unload tab panel respects the `pinnedNoUnload` setting

FurkanKambay:
Unloading a tabs panel by middle clicking it on the navbar always ignores pinned tabs, regardless of the `Prevent pinned tabs from unloading` setting, as opposed to using the context menu action and the `Unload all tabs in inactive panels` keybinding.

reproduce/verify with:
- `Prevent pinned tabs from unloading`: `off`
- Navigation bar actions > `Middle click on tabs panel`: `unload tabs`
- have 1 loaded pinned tab in a panel and discard the panel using MMB or context menu


mbnuqw:
Thanks

PR #2344, opened 2026-01-26
fix: keep "Prevent pinned tabs from unloading" setting enabled when pinned tabs are global

### LINKED ISSUE #2309: Globally pinned tab should not be unloaded
### Steps to reproduce

- Install Sidebery
- `Pinned tabs position` = `globally - top/left/right`
- `Prevent pinned tabs from unloading` is grayed out and left untouched (= `off`)

### Actual behavior

It is possible to unload pinned tabs via context menu.

### Expected behavior

It shouldn't.

Temporary workaround:
- `Pinned tabs position` = `in panel - top`
- `Prevent pinned tabs from unloading` = `on`
- `Pinned tabs position` = `globally - top/left/right`
- Now it behaves like it should.

### System

Windows 11/10

### Firefox version

145.0.1 (latest)

### Sidebery version

 5.4.0 (latest)

### Logs

```code

```

FurkanKambay:
Resolves #2309

also please see: https://github.com/mbnuqw/sidebery/issues/2309#issuecomment-3801133145

> Right clicking a specific pinned tab to unload it should bypass the setting maybe?
> And if not, then the unload option should be hidden in the context menu depending on the prevent unload setting's value.

### DIFF
--- a/src/page.setup/components/settings.tabs.vue
+++ b/src/page.setup/components/settings.tabs.vue
@@ -297,7 +297,6 @@ section(ref="el")
       @update:value="Settings.saveDebounced(150)")
     ToggleField(
       label="settings.pinned.no_unload"
-      :inactive="Settings.state.pinnedTabsPosition !== 'panel'"
       dbg="pinnedNoUnload"
       v-model:value="Settings.state.pinnedNoUnload"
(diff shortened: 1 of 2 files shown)


sonbua (approved):

mbnuqw:
Thanks, I'll add more configurability to this (for explicit unloading)

PR #2373, opened 2026-03-14
fix: restore recently closed tabs with navigation history

Draly:
## Summary

- Use `browser.sessions.restore()` instead of `Tabs.open()` when reopening tabs from the "Recently closed tabs" panel (trash icon), so the restored tab preserves its full navigation history (back/forward buttons work).
- Falls back to the previous `Tabs.open()` behavior when the tab is no longer available in browser sessions (e.g. older than 25 entries).
- Handles branch restoration: tries session restore for each tab individually, with fallback for tabs not found in sessions.

## Problem

Currently, clicking a tab in the "Recently closed tabs" sub-panel creates a brand new tab via `browser.tabs.create({ url })`. This means the restored tab loses its entire navigation history -- the back button doesn't work, unlike Firefox's native "Recently closed tabs" (History menu) which uses `browser.sessions.restore()`.

## Solution

Modified the `openTabs()` function in `sub-panel.closed-tabs.vue` to:
1. First attempt to find the closed tab in `browser.sessions.getRecentlyClosed()`
2. If found, restore it via `browser.sessions.restore(sessionId)` (preserves history)
3. If not found, fall back to the existing `Tabs.open()` behavior

### Limitations

- `browser.sessions.getRecentlyClosed()` is limited to ~25 entries by default (`browser.sessionstore.max_tabs_undo`). Tabs beyond this limit will use the fallback.
- `browser.sessions.restore()` restores the tab at its original position; Sidebery's tab event handlers will then integrate it into the tree.

## Test plan

- [ ] Open a tab, navigate through several pages (building history), close the tab
- [ ] Open the "Recently closed tabs" panel (trash icon) and click the closed tab
- [ ] Verify the back button works and navigation history is preserved
- [ ] Test with branch restoration (close a parent tab with children, restore via branch button)
- [ ] Test fallback: close many tabs (>25) and try restoring an old one -- should still open via URL

### DIFF
--- a/src/sidebar/components/sub-panel.closed-tabs.vue
+++ b/src/sidebar/components/sub-panel.closed-tabs.vue
@@ -43,6 +43,7 @@ import * as Sidebar from 'src/services/sidebar.fg'
 import * as DnD from 'src/services/drag-and-drop.fg'
 import * as Windows from 'src/services/windows.fg'
 import * as Utils from 'src/utils'
+import * as Logs from 'src/services/logs'
 import ScrollBox from 'src/components/scroll-box.vue'
 
 const state = reactive({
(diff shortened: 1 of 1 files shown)


mbnuqw:
Thanks. Currently there are some issues:
- Container is not preserved:
  - Open `https://example.com/` in default container
  - Open `https://example.com/` in another container
  - Close the tab in another container
  - Close the tab in default container
  - Open RCT sub-panel and restore containered tab
  - Restored tab is wrong (different container, history)

- Restoring branch of tabs leads to multiple active tabs:
  - Open `https://example.com/`
  - Open child tab `https://www.iana.org/help/example-domains` ("Learn more" link)
  - Close the branch (right-click on parent tab's close button)
  - Open RCT sub-panel and restore the branch
  - Both tabs marked as active

PR #2383, opened 2026-03-31
issue-2359: fix drag and drop one pixel before slot start case

### LINKED ISSUE #2359: Drag and drop on a group sometimes results in tree being moved to bottom of panel
### Steps to reproduce

- Install Sidebery
- Swiftly drag and drop tab onto a group

### Actual behavior

- Tab is sometimes moved to bottom of panel

### Expected behavior, possible causes and additional notes

- Tab goes in group it was dropped within

### System

Windows 11

### Firefox version

148.0b15

### Sidebery version

v5.4.0

### Logs

No logs because I can't repro consistently, happens intermittently with no pattern I've identified yet. Not super often.

the-nelsonator:
Fixes https://github.com/mbnuqw/sidebery/issues/2359

If group you're dragging a tab onto is at index 3 and then in `drag-and-drop.fg.ts#onDragMove`

```
  y = bounds[3].start - 1
```
DnD.reactive.dstParentId isn't set, so remains -1, so DnD results in tab being moved to bottom/top of panel depending on config.

Changes allows execution to make it to `// Inside` section when this happens. It handles as expected then.

I'm not sure if there's a more elegant fix or if this covers all cases, but this seems to fix this one particular case. If you can find a superset of cases feel free to discard for a better fix.

### DIFF
--- a/src/services/drag-and-drop.fg.ts
+++ b/src/services/drag-and-drop.fg.ts
@@ -915,7 +915,7 @@ export function onDragMove(e: DragEvent): void {
     if (y > slot.end) path[slot.lvl] = slot
 
     // Skip
-    if (!lvlChanged && (y > slot.end || y < slot.start)) continue
+    if (!lvlChanged && (y > slot.end || y < slot.start - 1)) continue
 
     // Between
     if (slot.in ? y < slot.top : y < slot.center) {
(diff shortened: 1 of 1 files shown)


mbnuqw:
Thanks. This PR has an issue: when slowly dragging a tab the pointer suddenly jumps to the next slot in "inside" mode.

mbnuqw:
And I still can repro the original issue.

mbnuqw:
Ok, I found out that a blind zone is not just a pixel but a range from slot.bottom to slot.end.

the-nelsonator:
Thanks for the proper fix! :)

Issue #1715, opened 2024-07-12
Option to have 'pinned tabs' act more like bookmarks


StudioNirin:
### Description

So the main reason I would pin a tab, is because that specific website link is one I want to keep going to (eg. Reddit.com) and so it is more like a bookmark than an active tab. At the moment pinned tabs are kind of useless, because if I open the tab and then browse to another page on it... the pinned tab is now some other page, and not the default I want. So I might as well just leave it in the main tabs list with everything else.

It would be nice to be able to add pinned 'bookmarks' instead, where you click on it and it opens a new tab in the tab list to that page. 

This works with the 'bookmarks panel' (if I change some of the default settings, the defaults are weird and break the bookmarks panel every time you click a link), but it means switching between the two panels. Would be nice to be able to add a small selection of pinned bookmarks to the top of the tabs panel.

wd:
+1
Arc has such a feature that I missed a lot. You can go back to the original URL by double-click on the tab.

awgneo:
This is the only feature holding me back from using Floorp+Sideberry entirely. I went through all of the options to try and replicate Arc's amazing tab functionality and Sideberry is 90% there. Sync to bookmarks, snapshots, these are all entirely unnecessary if one could "lock" a tab to an editable URL and, say by clicking the icon Arc-style, restore the tab to its home base. Don't get me wrong, I love backups, but locking tabs to a home base, bases that should be all restored optionally after re-opening the browser, would bring in the Arc-base. I'd almost say deprecate sync to bookmarks (manual process) and snapshots (manual restore), and just provide a locking method for any tab/group and a good backup option. If that is bookmarks, so be it. If that is snapshots, so be it. Also, I wanted to reference this bug https://github.com/mbnuqw/sidebery/issues/2028, which is really bad when trying to restore tabs from bookmarks to provide a sort-of-too-wholistic-locked-tab experience.

albino1:
Maybe I'm misunderstanding, but why would you go to a different website on a pinned tab?  Just use a new tab.  Your pinned tab should be only that website, that's the whole point of pinned tabs.  Using the original example, if you have a pinned reddit tab, then just never go anywhere on it that isn't reddit and it will always be reddit when you click on it.

[...]

Issue #2086, opened 2025-04-04
[Feature Request] Save tabs/pinned tabs state and ability to restore


DRYN07:
### Description

It would be great to have a solution that can save either each tabs or just the state (URL) of the pinned tabs, and restore to the default saved state after a shortcut or unload.

mikedlr:
My current workaround is to make a bookmark folder and then restore that into a tab. This doesn't work fully:

* doesn't restore as a pinned tab - it restores as a normal tab
* restores to the default container of the workspace, not a specific chosen container (unless there's a rule which forces that site to a particular container).
* restores as lazy loaded - no simple way to say these should all load directly.

however it does do a bunch of good things

* attempts to load exactly the sites you want and not every random site that you happened to have open at the moment firefox crashed because of one of those sites using too much memory
* you can remember which tabs and activate them
* it happens on demand rather than at browser startup which means that you get to use your browser immediately rather than waiting 20 minutes as you unload all sorts of buggy web sites that get unhappy about being restarted. 

so it's much much better than current firefox session restore. 

There are various tab restore extensions - haven't found one that even advertises sideberry compatibility yet though. Suggestions appreciated.

the-nelsonator:
I'm confused on the ask here. Snapshots save pinned tabs, so you can save them and restore them, just like normal tabs.

You want to have a set of pinned tabs that you can switch to via shortcut?

DRYN07:
Basically, I work with a lot of tabs, including many pinned tabs that are essential for each project, and I like to access them quickly. However, it often happens that I navigate back and forth in a project's drive throughout the day, accidentally switching to new URLs in the pinned tab, and it would be convenient to "reset" a specific tab to a saved state/URL with the press of a button. So, I don’t want to restore an entire snapshot because I’m continuously working on multiple projects, and I don’t aim to delete the tabs currently in progress, but I just want to restore individual tabs to their original position. It would already be a great help if all pinned tabs automatically reset after a restart. If there were a context menu option to save any tab in its current state (not just pinned), that would be perfect.

mikedlr:
@the-nelsonator snapshots work okay for me, however
* you normally have at least one tab or window that you don't want to restore (at least the window that you used to save the snapshot)
* you have to go back in the history to find your clean snapshot 

Being able to name a snapshot, select the particular tabs and/or windows you want then move it to the bookmarks bar and restore in one click would make it much more useful.

FurkanKambay:
[...]

Issue #1007, opened 2023-03-30
Support for Local Pinned tabs ( per panel basis ) and Global Pinned Tabs.


KAGEYAM4:
If there was additional option to have `local pinned tabs per panel basis` that would be awesome. So i can have Tabs that i want to share across all `panels` view ( Global Pinned Tabs )  and tabs that will be local to specific `panels` ( local pinned tabs).

microbearsmb:
This would be incredible

dnut:
edit: I wrote this comment with some incorrect assumptions. see below

We already have local pinned tabs, right? Pinned tabs, like normal tabs, are always confined to a panel. This feature request is to add support for globally pinned tabs.

Globally pinned tabs would be great. Typically I want my pinned tabs (email and calendar) to appear in every panel. I end up duplicating the same tabs everywhere, but their states get out of sync, and I get duplicate notifications, which is pretty annoying.

KAGEYAM4:
> We already have local pinned tabs, right? Pinned tabs, like normal tabs, are always confined to a panel. This feature request is to add support for globally pinned tabs.
> 
> Globally pinned tabs would be great. Typically I want my pinned tabs (email and calendar) to appear in every panel. I end up duplicating the same tabs everywhere, but their states get out of sync, and I get duplicate notifications, which is pretty annoying.

Currently we have both but they are mutually exclusive - if using local pinned then global pinned is disabled and if using global pinned then local pinned is disabled.

Having Both at same would be good.

microbearsmb:
We have Global pinned tabs? Awesome, didn't realize that. But yeah, having both would be so useful. I am a college student, so being able to both pin things like Spotify globally, and then course syllabi and gradebooks locally in each course's panel would be so nice.

KAGEYAM4:
> We have Global pinned tabs? Awesome, didn't realize that. But yeah, having both would be so useful. I am a college student, so being able to both pin things like Spotify globally, and then course syllabi and gradebooks locally in each course's panel would be so nice.

![image](https://github.com/mbnuqw/sidebery/assets/75798544/524c8e92-6d4d-46fc-91e3-a8dbffada1f2)

microbearsmb:
> > We have Global pinned tabs? Awesome, didn't realize that. But yeah, having both would be so useful. I am a college student, so being able to both pin things like Spotify globally, and then course syllabi and gradebooks locally in each course's panel would be so nice.
> 
[...]

PR #2384, opened 2026-03-31
feat: anchor tabs

VictorVow:
## Summary

Adds an **anchor tabs** feature. An anchored tab remembers its current URL and title as a "home" URL. When the user navigates the tab to a different URL, a configurable reset indicator appears between the favicon and the tab title. Hovering the favicon area shows a "Back to \<title\>" affordance and clicking it navigates back to the anchor URL. Unanchoring clears the saved data.

## Screenshots

Anchor tabs get a small permanent highlight around the favicon to show that it's anchored:

2026-03-31T16 22 55 453Z

When you navigate to a different URL the tab shows a "reset indicator". You can use whatever you want for this (defaults to `\`):

2026-03-31T16 20 51 801Z

Mouseover the highlighted area and the favicon will change icon, and you'll see "Back to \<title\>". Clicking this button will return the tab to it's anchor URL:

2026-03-31T16 26 33 360Z

## Behaviour

- **Anchor / Unanchor** — right-click context menu entry below Pin; also available as a keybinding (`toggle_anchor_tab`, no default shortcut)
- **Permanent highlight** — anchored tabs show a persistent highlight on the favicon area so they are visually distinct at a glance
- **Reset indicator** — configurable text/emoji indicator (default `\`) appears between the favicon and title when the tab has navigated away from its anchor URL
- **Hover to return** — hovering the highlighted area on an anchored-away tab swaps the favicon to an undo icon, shows "Back to \<title\>" in place of the tab title, and changes the cursor to a pointer; clicking anywhere in that area navigates back to the anchor URL
- **Persistence** — anchor URL and title are stored in `tabsDataCache` and restored across browser restarts

## Settings

Settings → Tabs → **Anchor tabs** — a text field to customise the reset indicator character (accepts text and emojis).

## Files changed

| Area | Files |
|---|---|
| Types | `src/types/tabs.ts`, `src/types/settings.ts` |
| Defaults | `src/defaults/settings.ts`, `src/defaults/menu.ts` |
| Services | `src/services/tabs.fg.ts`, `src/services/tabs.fg.handlers.ts`, `src/services/keybindings.fg.ts`, `src/services/menu.fg.options.tabs.ts`, `src/services/setup-page.fg.ts` |
| UI | `src/sidebar/components/tab.vue`, `src/styles/sidebar/tab.styl` |
| Settings page | `src/page.setup/components/settings.tabs.vue`, `src/page.setup/components/keybindings.vue` |
| Manifest / i18n | `src/manifest.json`, `src/_locales/dict.browser.json`, `src/_locales/dict.common.ts`, `src/_locales/dict.setup-page.ts` |
| Mocks | `src/defaults/mocks.tabs.fg.ts` |

## Closes
https://github.com/mbnuqw/sidebery/issues/1715
https://github.com/mbnuqw/sidebery/issues/2086


FurkanKambay:
A design consideration: having separate actions for 'anchor' and 'pin' might be undesirable and/or unexpected for some users, for these reasons:

1. Complexity: it's one more thing to keep track of regarding a tab's state, both in the codebase and from the user's perspective. Basically I think that it's not good UX.
2. Familiarity: they might be used to or expect the Zen/Arc behavior where all Pinned tabs are 'anchored'. That way is also more battle-tested and proven as a good design, in my opinion.

Personally, I'd prefer all pinned tabs be anchored to their initial URL and have actions to either "re-anchor to current URL" or "return to anchored URL".

VictorVow:
Fair point, I didn't want to alter such a well-known feature but I agree it may be better UX simply modifying it. Thoughts @mbnuqw?

mbnuqw:
Thanks, a few thoughts on this:

Naming: "Anchor/Unanchor" clashes with the terminology chosen in #1007. How about "Bind/Unbind URL" instead?

UI: Favicon area of tab is already used for "expand/fold" button in normal tabs. For pinned tabs, clicking should only activate the tab; changing the URL would be unexpected behavior for the user. Additionally, pinned tab has a clickable "audio" badge showed in its top-right corner so we can't just add another interactive element without increasing min-size of the pinned tab (which may happen in the future though). Anyway, unfortunately there is no vacant place for the new clickable button in the tab component right now. Technically it's possible but this will increase misclick rate or it will overlay other interactive zones, so I'd rather not.

We can only passively indicate tab state e.g. show something at the bottom-left corner of the favicon. And for example use some mouse actions (long/double/middle clicks) for fast bind/unbind/return. + keybinding and context menu opt for returning to the bound url.

VictorVow:
Implemented all the feedback, and the indicator now looks like this when a user is on the bound URL:
image

And this when the user has navigated to another URL:
image

Removed "hover to return", customisable reset indicator, permanent highlight. Renamed to "Bind/Unbind URL". Context menu or configurable shortcuts can be used for bind/unbind/return actions.

src/_locales/dict.common.ts
```diff
@@ -1933,6 +1933,39 @@ export const commonTranslations: Translations = {
     zh_TW: '取消釘選分頁',
     ja: 'ピン留め解除',
   },
+  'menu.tab.anchor': {
```
FurkanKambay: "anchor" usage here and in other places should be changed to not mix terminologies

src/_locales/dict.common.ts
```diff
@@ -1933,6 +1933,39 @@ export const commonTranslations: Translations = {
...
+    en: 'Bind URL',
+    de: 'Ankern',
+    fr: 'Ancrer',
+    hu: 'Horgonyzás',
+    pl: 'Zakotwicz',
+    ru: 'Привязать',
+    zh_CN: '锚定',
+    zh_TW: '錨定',
+    ja: 'アンカー設定',
```
FurkanKambay: I'd urge against using AI translations and suggest simply leaving most of these out unless you speak the language and are sure that these are correct in context. (_anchor_ probably would've been more problematic than _bind_ though)

The Japanese one, for instance, seems to mean "anchor settings" and not "anchor (this tab)" when back-translated into English and some other languages. Similar problems might occur with translations of bind/unbind without being this obvious.

Also, they all need to say "Bind URL" now

FurkanKambay (changes requested):

zelch:
Looked through the changes, and the only issues that I noticed had already been covered by @FurkanKambay.  All of the internal references to 'anchor' being the big one.

PR #2388, opened 2026-04-12
Feature/youtube buttons

bwhalley:
As described in a [discussion thread](https://github.com/mbnuqw/sidebery/discussions/2374). Adds a basic YouTube player interface to the bottom of the panel bar. The bar is defaulted off and controlled through a settings panel. I included a best try at some basic translations for the new UI elements. I tried to keep the code as tight as possible here, while supporting a range of YouTube player URLs. 
This could also be easily extended to any media player instead. 

image
image

### DIFF
--- a/src/_locales/dict.sidebar.ts
+++ b/src/_locales/dict.sidebar.ts
@@ -15,6 +15,55 @@ export const sidebarTranslations: Translations = {
     zh_TW: '搜尋...',
     ja: '検索...',
   },
+  'bar.youtube.no_playing': {
+    en: 'No YouTube tab is playing or paused here',
+    de: 'Kein YouTube-Tab spielt oder pausiert',
+    fr: 'Aucun onglet YouTube en lecture ou en pause',
+    ru: 'Нет вкладки YouTube с воспроизведением или паузой',
+    ja: '再生中・一時停止中の YouTube タブがありません',
+  },
+  'bar.youtube.prev': {
+    en: 'Previous video',
+    de: 'Vorheriges Video',
+    fr: 'Vidéo précédente',
+    ru: 'Предыдущее видео',
+    ja: '前の動画',
+  },
+  'bar.youtube.next': {
+    en: 'Next video',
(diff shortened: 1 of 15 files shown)


FurkanKambay:
looking forward to this one!

some things I'd love to see in the future for this:
- option to display the thumbnail + icon of website (to identify youtube, twitch, etc.)
- support multiple with a scroll UI (I imagine some dots on the right and vertical slide animations)
- pop-out player button

src/_locales/dict.sidebar.ts
```diff
@@ -15,6 +15,48 @@ export const sidebarTranslations: Translations = {
...
+    en: 'Pause',
+    de: 'Pause',
+    fr: 'Pause',
+    ru: 'Пауза',
+    ja: '一時停止',
+  },
+  'bar.youtube.mute': {
+    en: 'Mute tab',
+    de: 'Tab stummschalten',
+    fr: 'Couper le son de l’onglet',
```
FurkanKambay, line 49: Using YouTube's own translations would be much better for correctness and consistency.

Picking this one just as an example. YouTube's translation is "Désactiver le son". The addition of _de l'onglet_ is a separate decision, which might be worth discussing as well.

Missing languages should be easy to add too, but I'm not sure what the Chinese ones correspond to on YouTube's language list which has Simplified, Traditional, and Hong Kong

mbnuqw:
I like the idea of having playback controls at hand(mouse), but only in a more generic implementation: when Sidebery will get its own API, it should provide an interface for third-party addons (i.e. media-ctrl-providers or something). This way, users will be able to install needed "providers" for a desired service (youtube, spotify, bandcamp, etc...). Don't want to hard code them into Sidebery.

Alternatives and examples for possible UI:
- https://github.com/ungive/media-control-extension
- https://github.com/tnychn/media-controller-extension
- https://docs.zen-browser.app/user-manual/pip#media-player

I'll keep this PR open (if you don't mind) as a reference and reminder.

PR #2396, opened 2026-05-03
l10n: update French translation

ariasuni:
Mostly just translate untranslated stuff, but also fix a few strings so that everything makes sense.

- Correctly translate “global” to be able to refer to “globally pinned tabs” in another string
- For “sidebar”: use “barre latérale” everywhere, instead of “panneau latéral” as Firefox do, to avoid confusion with “panneau” (panel)


mbnuqw:
Thanks

PR #2441, opened 2026-08-09
perf: snapshots: use TextEncoder instead of new Blob to fix heavy GC lag

llc0930:
Tested for over a month in an environment with 9,000+ tabs and the Firefox profile on a traditional HDD, with a noticeable reduction in perceived lag when deleting snapshots.

### DIFF
--- a/src/services/snapshots.bg.ts
+++ b/src/services/snapshots.bg.ts
@@ -480,6 +480,9 @@ async function openWindow(
   await Windows.createWithTabs(items, { incognito: incognito })
 }
 
+const sizeCalcBuffer = new Uint8Array(MAX_SIZE_LIMIT * 1024 + 4)
+const sizeCalcEncoder = new TextEncoder()
+
 function limitSnapshots(snapshots: Snapshot[]): Snapshot[] | undefined {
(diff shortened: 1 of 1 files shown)


mbnuqw:
I profiled both methods (`Blob` and `TextEncoder`) with a non-ASCII character set (Zh): Both methods produced approx the same memory footprints. Both methods triggered string re-encoding (x2 CPU usage compared to ASCII-only chars) + `memcpy`. And they showed more or less the same GC behavior. The `Blob` method has a bit lower memory threshold that triggers GC, which probably causes this effect. But if the data size continue to grow, you'll start experiencing heavy GC lag with `TextEncoder` too.

To fix this we can pre-allocate and keep some memory for measuring (e.g. 10mb). But it brings two trade-offs:
- size measurement will be strictly limited to this 10mb, which is ok, there is a hard limit of 10mb on all snapshots anyway (or we could dynamically re-create this buffer, but this will trigger GC...).
- this 10mb of RAM will always be consumed by Sidebery. I think we can afford such luxury even these days.

btw, this size measurement will be removed when I migrate snapshots storage to OPFS.

PR #2442, opened 2026-08-09
feat: support public suffix matching for domain handling

llc0930:
This PR introduces public suffix list (PSL) matching to improve domain handling, addressing the discussion outlined in https://github.com/w3c/webextensions/issues/231#issuecomment-5204385729


mbnuqw:
While the `publicSuffix` API is actually useful in some parts of Sidebery (including the legacy `clearTabsCookies`), using it in wrongly named `Utils.getDomainOf` (it should be renamed to getHostname) is wrong. Another problem is using heavy `new URL()`. The `Utils.getDomainOf` is used in hot paths and should rely on lightweight regexp.exec (~20x faster).

PR #2443, opened 2026-08-11
fix: use moveNewTab setting for tab position when moving by rule

codogeno:
Previously, moveTabToPanel() used moveNewTabParent to determine the position of a tab in the target panel when moved by an auto-move rule. This caused tabs to always land at the end of the panel, ignoring the general new tab position setting (moveNewTab).

Now moveTabToPanel() calls getIndexForNewTab() with an empty config, which correctly applies the moveNewTab setting — the same logic used when creating a new tab without a parent.

Add tests covering start/end placement and verifying that moveNewTabParent no longer affects rule-based moves.


mbnuqw:
It should be a new specific setting for auto-moved tabs. `moveNewTab` is for new (just created) tabs only.

